PR #6, opened 2022-01-12
Update MAGIC-LST scripts to ctapipe v0.12.0

YoshikiOhtani:
Here the combined analysis scripts are updated to be used in the ctapipe v0.12.0 environment.


magicctapipe/scripts/lst1_magic_real/data/config.yaml
```diff
@@ -43,41 +73,82 @@ event_coincidence:
     
 
 stereo_reco: 
-    subarray: '/fefs/home/yoshiki.ohtani/software/magic-cta-pipe/lst1_magic/data/subarray_lst1_magic.pkl'
-    quality_cuts: (intensity > 50) & (width > 0)
+    subarray: '/home/yoshiki.ohtani/software/magic-cta-pipe/magicctapipe/scripts/lst1_magic_real/data/subarray_lst1_magic.pickle'
```
aleberti, line 76: I think that is better to save the subarray information together with the DL1 file(s) produced by the event coincidence script. If you have the SubarrayDescription object (as saved in the pickle file), you can define it in the event coincidence script and use the SubarrayDescription.to_hdf method to write it into the HDF5 file (if I understood correctly, saving the subarray in the HDF5 file should be done before saving anything else). 
This would avoid having to specify the path to the pickle file, and instead reading everything from the file at once.

magicctapipe/scripts/lst1_magic_real/lst1_magic_mc_dl0_to_dl1.py
```diff
@@ -1,248 +1,402 @@
...
+    mc_disp = Field(-1, 'MC event disp', u.deg)
+    mc_core_x = Field(-1, 'MC core x', u.m)
+    mc_core_y = Field(-1, 'MC core y', u.m)
+    mc_impact = Field(-1, 'MC impact', u.m)
     n_islands = Field(-1, 'Number of image islands')
+    n_pixels = Field(-1, 'Number of pixels of cleaned images')
+    magic_stereo = Field(-1, 'True if M1 and M2 are triggered')
+
+
+class SimInfoContainer(Container):
```
aleberti, line 78: Why not using the  SimulationConfigContainer from ctapipe instead of defining a new one?  SimulationConfigContainer has more members, so you will not use them all, but I think we should try to use already defined things (especially if already in ctapipe)

aleberti: Ok, now I see that you extract `source.simulation_config`, which indeed is a `SimulationConfigContainer`, and then fill the custom container. Since the simulation configuration is not a big container (anyway just one per run), I would write directly `source.simulation_config` into the HDF5 file.

YoshikiOhtani: Thanks for your comment, and indeed I agree with you, I will change it.

magicctapipe/scripts/lst1_magic_real/lst1_magic_stereo_reco.py
```diff
@@ -162,75 +211,76 @@ def stereo_reco(input_data, output_data, config):
...
-            array_pointing
-        )
+            event.dl1.tel[tel_id].parameters = ImageParametersContainer(hillas=hillas_params)
 
+        hillas_reconstructor(event)
+        stereo_params = event.dl2.stereo.geometry["HillasReconstructor"]
+            
         if stereo_params.az < 0:
-            stereo_params.az = stereo_params.az + u.Quantity(2*np.pi, u.rad)
+                stereo_params.az += u.Quantity(360, u.deg)
```
aleberti, line 220: here the indentation is wrong, should be one tab less

YoshikiOhtani: thanks I modified it

magicctapipe/scripts/lst1_magic_real/data/config.yaml
YoshikiOhtani: Hm, OK, I think I can define the subarray object inside the stereo reconstruction script. I'm implementing this solution now.

setup.py
```diff
@@ -22,11 +22,11 @@
     install_requires=[
         'astropy>=4.0.5,<5',
         'ctapipe~=0.12.0',
-        'ctapipe_io_magic~=0.3.0',
+        'ctapipe_io_magic',
```
aleberti, line 25: Was this an issue for you? Anyway this will be changed when we will merge into master and I will make a new version of ctapipe_io_magic (v0.4.0, which I will create from the current master branch, the one working with ctapipe v0.12)

aleberti (changes requested):
Ok, I think I checked everything, and put my comments, and some you already replied. So at that point I will check them again. Thanks for updating the scripts!

setup.py
YoshikiOhtani: Yes there was an issue for some reason, anyway the version 0.3.0 is for ctapipe v0.8.0 and so I modified it. Would be great if you create v0.4.0, and then I will modify it again.

aleberti: Ok, I can do it now. I will check if I need to update a bit the README.md and then create tag and release.

magicctapipe/scripts/lst1_magic_real/data/config.yaml
YoshikiOhtani: By the way saving the subarray after saving others, for example event dl1 parameters works without any problems. Probably it was updated?

setup.py
aleberti: Ok, it took a bit, but now ctapipe_io_magic was tagged with version 0.4.0. Also, it can be installed with pip without cloning the repo, since now I uploaded the module in PyPI.

YoshikiOhtani: Thanks! Ah so it means now one doesn't need to clone both ctapipe and ctapipe_io_magic, and instead just need to create an environment, clone only magic-cta-pipe and type `pip install .`?

YoshikiOhtani: Hm, actually I succeeded in installing all the software with the following steps:
$ conda env create -n magic-cta-pipe
$ git clone https://github.com/cta-observatory/magic-cta-pipe.git 
(after changing the setpy.py about the ctapipe_io_magic dependency)
$ pip install ./magic-cta-pipe/
But probably it would be better to prepare a environment.yml file to create an environment as it is done by ctapipe or lstchain?

aleberti: To have everything working as you say, indeed we would need to create an environment.yml file, which will install ctapipe 0.12 (and dependencies), ctapipe_io_magic 0.4.0 (and dependencies) and finally magic-cta-pipe

YoshikiOhtani: I see, is it possible for you to create an environment.yml file? I think it would be better and easy way for the students to install the pipeline if they need to do it by themself (in the local machine, another computer, etc)

aleberti: Sure, I can do it. I will do a commit in the update-ctapipe-v0.12 branch.

aleberti: Done!

YoshikiOhtani:
Sorry for taking time but I found some mistakes in the scripts about the subarray descriptions, and so I modified them. I also updated the script for merging HDF files. I tested the script and confirmed that now it works fine.

YoshikiOhtani:
@aleberti, if it is fine for you, could you merge this branch to update-ctapipe-v0.12, and merge update-ctapipe-v0.12 to master, please? And probably create a new tag (v0.1.0 ?) for the current version?

aleberti (approved):

aleberti:
Sure, give me some time and I do it :)

YoshikiOhtani:
Thanks a lot! I will delete my branch then…

PR #12, opened 2022-01-17
Refactor MAGIC bad pixels calculator

### LINKED ISSUE #11: Refactor MAGIC_Badpixels.py
Given that ctapipe_io_magic was refactored, I think some modules used in this repository can be also refactored. I looked through the `MAGIC_Badpixels.py` and have some ideas of the refactoring:

- it would be better to use the dictionary type for the output of the get_{badrmspixel, deadpixel}_mask function with the keys of the telescope IDs (M1: 1, M2: 2). This makes the Cal->DL1 script more readable. Currently the list type is returned and for example M1 mask is extracted as `mask[tel_id - 1] `. 
- since we now input only one subrun file, the observation ID will not change in a given run and so the `_check_new_run` function is not needed anymore.
- the `_getpixratiosqrt` just returns 1 and so is not needed anymore.

YoshikiOhtani:
This pull request contains the refactoring of the MAGIC bad pixels calculator. Closes #11. It is still in progress and so will inform you once it is ready for reviews.

### DIFF
--- a/magicctapipe/utils/MAGIC_Badpixels.py
+++ b/magicctapipe/utils/MAGIC_Badpixels.py
@@ -6,7 +6,7 @@ __all__ = [
 ]
 
 
-class MAGICBadPixelsCalc():
+class MAGICBadPixelsCalc:
 
     def __init__(self, is_simulation, camera=None, config=None, tool=None, **kwargs):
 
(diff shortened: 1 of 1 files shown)


YoshikiOhtani:
Since I think the bad pixels treatment should be done inside the ctapipe_io_magic let me close this pull request and open a new one for ctapipe_io_magic.

PR #16, opened 2022-01-18
Console scripts

### LINKED ISSUE #15: Rename scripts directory
Is is OK for you @aleberti to rename the directories of the LST-1 + MAGIC analysis scripts, because the `lst1_magic_real` directory contains not only for real data analysis scripts but also the MC data analysis scripts (`lst1_magic_mc_dl0_to_dl1.py`, `lst1_magic_train_rfs.py`, `lst1_magic_dl1_to_dl2.py`) and so the current name is a bit strange. Probably, the `lst1_magic_mc` can be renamed something like `sensitivity` or `performance`, and then the `lst1_magic_real` can be renamed `lst1_magic`. What do you think?

aleberti:
Add entry points to allow use of command line scripts for MAGIC+LST1 analysis. Renaming of folders. Closes #15 .


YoshikiOhtani:
Thank you very much for your modifications, much appreciated!

PR #18, opened 2022-02-01
Reorganize the repository 

YoshikiOhtani:
Let me reorganize the structure of the repository


YoshikiOhtani:
So what I mainly did is:
- renamed the directory, from "reco" to "image", and from "train" to "reco"
- removed MAGIC_Badpixels.py because the function will be implemented in ctapipe_io_magic
- renamed the class of MAGIC_Cleaning.py to the capital ones
- Moved some modules in utils directory to image or reco directory

I have to also modify the scripts because the modules are moved to the different directory, which will be done tomorrow. Anyway any comments are welcome.

aleberti:
I will add some commits to this branch, updating some of the scripts in the `magic` directory. In this way I will update them and also have two versions of the same scripts (one from this branch, one from the master) in order to cross-check the new bad pixels implementation in `ctapipe-io-magic`

YoshikiOhtani:
Ah sorry I just rebased this branch - actually I deleted the commit to remove the script of the bad pixel calculation for the cross-check purpose. If you are fine, @aleberti, can I merge this to the master branch? Then you can also merge your development #5 to master, and can cross-check the bad pixels calculation in master branch?

aleberti:
Ok, I don't see the reason to merge into master now, but you can do it. Anyway, #5 is not for comparing images between different MCP versions (as I need now), but images between MARS and MCP (same for hillas parameters). Once you merge into master, I will create another branch to commit the modifications I need.

YoshikiOhtani:
OK thanks, the reason is that I just don't want to keep this branch so long time because it just reorganizes the structure, and am doing some development starting from this branch.

PR #29, opened 2022-02-25
Create DL3 files with pyirf and gammapy

YoshikiOhtani:
This pull request includes additional scripts to create the IRFs and DL3 files with pyirf and gammapy. In addition, the already-existing scripts will be updated as following the requests from pyirf.


notebooks/check_sed_from_dl3.ipynb
```diff
@@ -0,0 +1,837 @@
+{
```
gabemery, line 1: <p>Do you now why there is this message : "MapMaker: use_region_center=False should not be used with point-like IRF. Results are likely inaccurate." ?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/magic-cta-pipe/pull/29/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

gabemery, line 1: <p>Ok, but maybe add a comment that plot_spectrum() defaults to e2dnde and erg cm-2 s-1 for the sed type and unit.</p><p>What you do is correct but may be confusing if this behaviour is not known (as it was my case before looking at the source code of gammapy)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/cta-observatory/magic-cta-pipe/pull/29/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

YoshikiOhtani: <p>No I don't... I see that the use_region_center option appears in the SpectrumDatasetMaker starting from gammapy v0.19:</p><p>https://docs.gammapy.org/0.19/api/gammapy.makers.SpectrumDatasetMaker.html?highlight=spectrumdatasetmaker#gammapy.makers.SpectrumDatasetMaker</p><p> </p><p>It's strange that the default value is True, but the message seems to indicate that the option is set to False. Anyway it would be better to explicitly specify the option to True?</p>

YoshikiOhtani: <p>OK, I will add the comment, thanks!</p><p> </p>

gabemery: I don't know. You could try to test if it changes anything, or look more at the code to understand what is happening.

YoshikiOhtani: <p>I checked it and found that the <code>use_region_center</code> is set to True as default, and of course the result doesn't change when I explicitly set the option to True. And I found the following line where this warning comes from, and this message appears when the input is point-like, even if the <code>use_region_center</code> option is set to True. So it looks like just a reminder, and since I actually set it to True, I think I just follow the correct procedure for point-like data.  </p><p> </p><p>https://github.com/gammapy/gammapy/blob/7e2976713facc0d443ade5279e91c4ae6b38c6c4/gammapy/makers/spectrum.py#L69</p><p> </p><p>Just for your information, If I set the option to False the process does not work. It shows an error message when running the following line, and it is because the <code>make_exposure</code> function doesn't work for point-like data if the option is set to False:</p><p> </p><p><code>dataset = dataset_maker.run(dataset_empty.copy(name=str(obs_id)), observation)</code></p>

gabemery: Ok so no problem on your side. I think this should be fixed in gammapy to only appear when `use_region_center` is True.
Actually it is already fixed in master [here](https://github.com/gammapy/gammapy/blob/master/gammapy/makers/spectrum.py#L69)

YoshikiOhtani:
Thank you Gabriel for your reviews and I implemented your comments. If there are any further comments, I will merge this branch to the master by the end of this month. It would be great if you could post your comments by then.

PR #34, opened 2022-03-14
Hillas image test

ellijobst:
Adding tests and scripts for image comparison and hillas parameter comparison


magicctapipe/scripts/magic/tests/params_for_hillas_comparison.py
```diff
@@ -0,0 +1,100 @@
...
+	"delta_M1" : "psi", "delta_M2" : "psi", 
+	"cogx_M1" : "x", "cogx_M2" : "x", 
+	"cogy_M1" : "y", "cogy_M2" : "y", 
+	# "hmax" : "h_max", 
+	# "corex" : "core_x", 
+	# "corey" : "core_y", 
+	# "az" : "az", 
+	# "zd" : "alt",
+	# "n_islands" : "n_islands",
+	"leakage_M1" : "intensity_width_1", "leakage_M2" : "intensity_width_2"
```
jsitarek, line 33: I guess there is a type with intensity_width_2 should be also intensity_width_1

ellijobst: Thanks for noticing. It is fixed now!

magicctapipe/scripts/mars/tests/compare_hillas_stereo_params.py
```diff
@@ -0,0 +1,291 @@
...
+    mcp_file = config["magic-cta-pipe-input"]["MCP-path"]
+
+    df_mcp = pd.read_hdf(mcp_file, key=params_key)
+
+    # filter for subrun
+    # df_mcp = df_mcp.loc[df_mcp["obs_id"] == subrun]
+
+    print(f"Number of events before cuts {df_mcp.shape[0]}")
+
+    # LON_ORM = u.Quantity(-17.89064, u.deg)
```
jsitarek, line 51: there is a lot of commented code here, will it be used for the further developments?

magicctapipe/scripts/mars/tests/compare_hillas_stereo_params.py
```diff
@@ -0,0 +1,291 @@
...
+            df_mars["mars_" + par] = file_mars["Events"][mars_name].array(library="np")
+
+    # apply cuts
+    # 50<Size<50000, leakage<0.15, impact<120
+    df_mars_cut = df_mars.loc[
+        (df_mars["mars_size_M1"] < 50000)
+        & (df_mars["mars_size_M2"] < 50000)
+        & (df_mars["mars_size_M1"] > 50)
+        & (df_mars["mars_size_M2"] > 50)
+        & (df_mars["mars_leakage_M1"] < 0.15)
```
jsitarek, line 124: leakage cuts are applied in MARS in the training, but the events are not removed later on in the  analysis (nevertheless often will have a high hadronness and thus would not survive the cuts)

magicctapipe/scripts/mars/tests/labels_and_units_for_params_comparison.py
```diff
@@ -0,0 +1,145 @@
...
+    "leakage_M1" : "MNewImagePar_1./MNewImagePar_1.fLeakage1", "leakage_M2" : "MNewImagePar_2./MNewImagePar_2.fLeakage1"
+
+}
+mcp_params = {
+    "length_M1" : "length", "length_M2" : "length",
+    "width_M1" : "width", "width_M2" : "width",
+    "size_M1" : "intensity", "size_M2" : "intensity",
+    "slope_M1" : "slope", "slope_M2" : "slope",
+    "delta_M1" : "psi", "delta_M2" : "psi",
+    "cogx_M1" : "y", "cogx_M2" : "y",
```
jsitarek, line 34: I think it would be good to add here a comment about the difference of the camera frame definition between MARS and ctapipe

magicctapipe/scripts/mars/tests/labels_and_units_for_params_comparison.py
```diff
@@ -0,0 +1,145 @@
...
+    "impact_disp_M1" : "impact", "impact_disp_M2" : "impact",
+    # "n_islands" : "n_islands",
+    "leakage_M1" : "intensity_width_1", "leakage_M2" : "intensity_width_1"
+}
+
+scale_factors = {
+    "length_M1" : 1000. / 1.0713, "length_M2" : 1000. / 1.0713,
+    "width_M1" : 1000. / 1.0713, "width_M2" : 1000. / 1.0713,
+    "size_M1" : 1., "size_M2" : 1.,
+    "slope_M1" : 0.0033703*1.64*1.0713, "slope_M2" : 0.0033703*1.64*1.0713,
```
jsitarek, line 56: I added the scaling for slope how I think it should be, but it actually does not work, and I am not sure why.
0.0033703/1.64/1.0713 would work, but it does not make sense to me why it should be like this

magicctapipe/scripts/mars/tests/labels_and_units_for_params_comparison.py
```diff
@@ -0,0 +1,145 @@
...
+    # "n_islands" : 1.
+}
+
+# important for the plotting, these are mars units
+labels_and_units = {
+    "length_M1" : "Length M1 [mm]", "length_M2" : "Length M2 [mm]",
+    "width_M1" : "Width M1 [mm]", "width_M2" : "Width M2 [mm]",
+    "size_M1" : "Size M1 [phe]", "size_M2" : "Size M2 [phe]",
+    "delta_M1" : "Delta M1 [rad]", "delta_M2" : "Delta M2 [rad]",
+    "slope_M1" : "Time Gradient M1", "slope_M2" : "Time Gradient M2",
```
jsitarek, line 82: units in MARS are time slices / mm

magicctapipe/scripts/mars/tests/labels_and_units_for_params_comparison.py
```diff
@@ -0,0 +1,145 @@
...
+    "zd" : "MStereoPar.fDirectionZd",
+    "impact_M1" : "MStereoPar.fM1Impact", "impact_M2" : "MStereoPar.fM2Impact",
+    "hmax_disp" : "MStereoParDisp.fMaxHeight",
+    "corex_disp" : "MStereoParDisp.fCoreX",
+    "corey_disp" : "MStereoParDisp.fCoreY",
+    "az_disp" : "MStereoParDisp.fDirectionAz",
+    "zd_disp" : "MStereoParDisp.fDirectionZd",
+    "impact_disp_M1" : "MStereoParDisp.fM1Impact", "impact_disp_M2" : "MStereoParDisp.fM2Impact",
+    # "n_islands" : "MImagePar_1./MImagePar_1.fNumIslands",
+    "leakage_M1" : "MNewImagePar_1./MNewImagePar_1.fLeakage1", "leakage_M2" : "MNewImagePar_2./MNewImagePar_2.fLeakage1"
```
jsitarek, line 25: why '/' ?

magicctapipe/scripts/mars/tests/test_compare_params.py
```diff
@@ -0,0 +1,223 @@
...
+
+    config_params_file = str(tmp_path / "compare_params_config.yaml")
+
+    with open(config_params_file, "w") as outfile:
+        yaml.dump(config_params, outfile, default_flow_style=False)
+
+    config_mcp = {
+        "mc_tel_ids": {"LST-1": 1, "MAGIC-I": 2, "MAGIC-II": 3},
+        "MAGIC": {
+            "image_extractor": {
```
jsitarek, line 147: is the image_extractor and charge_correction needed here? the MARS files are already calibrated and in MCP this part is only needed for sim_telarray files processing

magicctapipe/scripts/mars/tests/test_compare_params.py
```diff
@@ -0,0 +1,223 @@
...
+                "use_time": True,
+                "use_sum": True,
+                "picture_thresh": 6,
+                "boundary_thresh": 3.5,
+                "max_time_off": 4.5,
+                "max_time_diff": 1.5,
+                "find_hotpixels": "auto",
+                "pedestal_type": "from_extractor_rndm",
+            },  # select 'fundamental', 'from_extractor' or 'from_extractor_rndm'
+            "muon_ring": {
```
jsitarek, line 163: I guess muon_ring can be also skipped

magicctapipe/scripts/mars/tests/test_compare_params.py
```diff
@@ -0,0 +1,223 @@
...
+                "thr_low": 25,
+                "tailcut": [12, 8],
+                "ring_completeness_threshold": 25,
+            },
+        },
+        "event_coincidence": {
```
jsitarek, line 169: this part should be only needed for MAGIC+LST1 joining,
MAGIC alone is joined by stereo event number

jsitarek:
Hi @aleberti @ellijobst 
I had a look at the part of comparing image parameters. I found a few bugs, but even after fixing them still many of the parameters fail the consistency tests. Was it like this before, or it only happened after merging the newest changes from master?

magicctapipe/scripts/mars/tests/compare_hillas_stereo_params.py
ellijobst: This part of the code is used to apply a theta2 cut when comparing the impact. But when applied only few events survived. Maybe it is better to just not use it at all?

magicctapipe/scripts/mars/tests/labels_and_units_for_params_comparison.py
ellijobst: Hi Julian, these factors are used to convert the units used in ctapipe into the MARS units. For the other parameters we also divide by e.g. 1.0713 so it would make sense to me, that it works if you divide instead of multiplying. But that is just a guess, I would need to check, since slope was the parameter that was still failing when we ran the tests.

magicctapipe/scripts/mars/tests/compare_hillas_stereo_params.py
jsitarek: for comparisons on the data such a cut would indeed remove nearly all the events so it does not make much sense. one could keep it for the MCs, but only if we really some a problem, i.e. impact matches nicely for closeby events, but ends up with much larger spread between both methods for high theta events (and actually we also have two 'theta' values - from MARS and from MCP that could be different as well)

magicctapipe/scripts/mars/tests/labels_and_units_for_params_comparison.py
jsitarek: yes, but e.g. length and width are in MARS in units of mm, while slope is time slices / mm, so if you divide the first one it is expected that the second one should be multiplied. Either way this 7% effect is not big, and here it is difficult to say if it matches better or not. I meant more the 1.64 factor. MARS files have slope in time slices / mm, and arrival times are given in time slices in the images, but then ctapipe_io_magic converts them to ns by dividing by 1.64 [GHZ].  We had to do it, because there was a problem when comparing with sim_telarray MCs, so what gets out of MCP should be in ns/deg. Multiplication by 0.00337 is making it .../mm and this part seems to work (not sure about the 1.07 factor), but to make it back time slices / mm it would need to be multiplied again by 1.64 (cancelling the original division), and this does not seem to work.

ellijobst:
Hi Julian,
we checked and, as you said, a lot more parameters failed the test than before. We think this is due to the fact that for this test we use smaller files with fewer events than before. So the thresholds(percentages of events that are allowed to have a difference) are too strict and the test fails. 
e.g. for the real data with a threshold of 0.5% it works as it is supposed to, but the simulated still fails. for it to work we would need to put very high thresholds, since it is maybe 1 or 2 events out of ~200. But this would really mess up the test for the real data. 
So we have two options: just use a high percentage and ignore the results for real data, or using two different percentages for real and simulated data. What do you think would be better?

jsitarek:
Hi @ellijobst,

I think we should try to understand what exactly is happening there. If I remember correctly I was getting ~8 events that were failing the test for MCs, and with the small MC file statistics this is really a large fraction. 
Please note also that one of the change I made will allow catching more failures than before (there was a problem that negative differences were ignored). 
For the data the very first subrun of each run might be special, because it is still using the calibration and pedestals from the calibration run - this in principle should only affect this pixels are selected as noisy and interpolated, so depending how much the MARS behaviour is mimicked it can also produce some changes. 

BTW, I was also trying with the image comparison and it fails completely, the signals do not match at all the references, not sure if this is the problem with the actual code or the comparison files.

magicctapipe/scripts/mars/tests/labels_and_units_for_params_comparison.py
aleberti: Hi Julian, are you sure you need the 0.00337 factor? As I see it, both hillas and timing parameters come out in unit of meters, and not degrees. They would be in degrees if the geometry is converted in the telescope frame (if I remember correctly, or camera one, now I am not sure) using the choice of focal length done at the EventSource level (i.e. nominal or effective), so that also the geometry units are degrees and all the parameters calculated starting from it.

aleberti:
> BTW, I was also trying with the image comparison and it fails completely, the signals do not match at all the references, not sure if this is the problem with the actual code or the comparison files.

about the images, it seems that MCP is sometimes selecting more pixels in the final cleaned image, typically at the border of the image. So this may point to something not correct when choosing the neighboring pixels. I attach an example that comes out from the tests. Of course in this case the difference in pixel charge for those pixels is not a fraction of p.e., but whatever the charge was in the calibrated image before cleaning, since simply those pixels are not selected in the final cleaned image in MARS.

image-comparison-824319_5_M2

ellijobst:
Hi Julian,

we fixed quite a bit of issues and now our results look a lot better than before. 

For the simulated data we still have a minor problem. For some events the charge of pixel no. 8 is interpolated in MARS and kept the same in MCP. Simulated data should not have bad pixels, therefore we are a bit confused about what is happening there. Maybe you have an idea?

aleberti:
Hi, adding a few details.

- the huge differences in images which we were seeing were mostly due to a wrong setting in the cleaning. Apparently we still had time cuts in time slices and not in nanoseconds (the cleaning configuration was not taken from the "official" configuration file, but hardcoded in `image_comparison.py` 
- we added the precision reduction as done in MARS, which acts when pixel charges are interpolated for bad pixels

So after this, I tested on the real data test files, without any limit on the number of events compared. Each of the 4 files (2 for M1 and 2 for M2), there are 500 events (this is because they are trimmed files), 2000 in total. Of these, 468 were skipped, because they are not present in the files containing the images saved from MARS (i.e. they did not pass the cleaning). Of the 1532 remaining events, 11 had slightly different images (7 with only one pixel, 1 with 2 pixels, 2 with 3 pixels and 1 with 11 pixels). Therefore we are at the level of ~0.7% of events with difference in the images. Actually I did not check what is the difference in charge in those pixels between MARS and MCP.

Now coming to the MC data. As Elli explained, for M1 some events have differences only for pixel id 8, or sometimes with more pixels (but id 8 is there always). So I added in the MARS code to print the pixels that are interpolated. Surprisingly, for the runs I took, and only M1, I have two bad pixels which are interpolated: 0 (which is a known "feature"), and 8. However when I try to get the bad pixel info from the root file like I do for real data, only 0 is tagged as bad. So what I tried to force the interpolation for MC data as well, where for M1 I interpolate pixels 0 and 8, and for M2 I interpolated only pixel 0. I run the test over all events in each file, and I got no differences. While I expected this result, since I am reproducing exactly what is done in MARS, still the question remains why this pixel 8 is tagged as bad in M1 only.

So that's the situation. Please let us know your opinion!

aleberti:
Here the difference in charge for the pixels:

```
[223]
Charge differences: [3.7109375]
errors found for 3823!
[205]
Charge differences: [3.390625]
errors found for 4012!
[361]
Charge differences: [7.09375]
errors found for 4055!
[625 626 714]
Charge differences: [17.0625    6.078125  3.84375 ]
errors found for 4156!
[265]
Charge differences: [5.71875]
errors found for 20726!
Charge differences: [ 5.84375    6.3125    14.1875     7.421875   8.25       4.953125
  8.5        7.375      5.578125   3.8359375  7.6875   ]
errors found for 20935!
[173 221 222]
Charge differences: [4.6875   4.359375 6.578125]
errors found for 21131!
[767]
Charge differences: [8.34375]
errors found for 4122!
[767]
Charge differences: [8.34375]
errors found for 4122!
[27 48]
Charge differences: [ 6.203125 12.25    ]
errors found for 20421!
[373]
Charge differences: [3.2578125]
errors found for 20636!
```

jsitarek:
Hi,

In MCs you can also get bad pixels if they end up as outliers by pure statistics. This might be the case for pixel 8. Since there are no interleaveds, bad pixels are assigned only on MC calibration run which is the same for all MC simulation runs, so such a pixel would appear bad in all the files of a given production. Do you have a scalib file corresponding to those MC files?

magicctapipe/scripts/mars/tests/labels_and_units_for_params_comparison.py
jsitarek: Sorry for a super late feedback on this. 
In the test that I did without the 0.00337 factor the slope was mismatching by a large factor. Maybe some other changes in the code affected it. Either way, at least 1.64 factor should be there, because MARS is using time slices and ctapipe ns. 
Are the tests working fine now for the slope parameter?

aleberti:
> Hi,
> 
> In MCs you can also get bad pixels if they end up as outliers by pure statistics. This might be the case for pixel 8. Since there are no interleaveds, bad pixels are assigned only on MC calibration run which is the same for all MC simulation runs, so such a pixel would appear bad in all the files of a given production. Do you have a scalib file corresponding to those MC files?

I did not find any scalib file. But I checked in the MC ROOT files, and within the `Events` tree I see that there are three branches `MPedPhotFundamental`, `MPedPhotFromExtractor` and `MPedPhotFromExtractorRndm`, which usually for real data are under the `Pedestals` tree. So I guess that outliers are calculated in a similar way to real data. I will update the MAGIC reader to read in this information and get "hot" pixels.

ellijobst:
Hi Julian,

I ran the image comparison over these files in the "/remote/ceph/group/magic/MAGIC-LST/Data/MAGIC/CrabNebula/MCP/image_comparison" directory:

"20220306_M1_05101249.001_Y_CrabNebula-W0.40+035.root",
"20220306_M1_05101249.002_Y_CrabNebula-W0.40+035.root",
"20220306_M2_05101249.001_Y_CrabNebula-W0.40+035.root",
"20220306_M2_05101249.002_Y_CrabNebula-W0.40+035.root",

"20220306_M1_05101249.001_I_CrabNebula-W0.40+035.h5",
"20220306_M1_05101249.002_I_CrabNebula-W0.40+035.h5",
"20220306_M2_05101249.001_I_CrabNebula-W0.40+035.h5",
"20220306_M2_05101249.002_I_CrabNebula-W0.40+035.h5",

for 500 events per file. So in total a bit less than 2000 events were compared.
I got these results:

FAILED test_image_comparison.py::test_image_comparison[dataset_calibrated4-dataset_images4] - 
assert [1465, 1501, 1551] == []
FAILED test_image_comparison.py::test_image_comparison[dataset_calibrated5-dataset_images5] - 
assert [14575] == []
FAILED test_image_comparison.py::test_image_comparison[dataset_calibrated6-dataset_images6] - 
assert [1304, 1814] == []
FAILED test_image_comparison.py::test_image_comparison[dataset_calibrated7-dataset_images7] - 
assert [14186, 14321, 14332, 14501] == []

So ~10/2000  events failed the test. I checked the images and for most of them it were just 1 or 2 pixels that were different.

Maybe we can put some threshold, so that the test does not fail if it is just a certain percentage of events that have an error (similar to the way it is done for the hillas/stereo params comparison). What percentage would you suggest?

aleberti:
So, if we compare images using time slices instead of time in ns, comparing 50000 events, I get 8 events with differences (mostly 1 or 3 pixels different), so 0.016%. Is this acceptable @jsitarek? Also, I made image comparison faster, 50000 events compared in less than 10minutes.

jsitarek:
> So, if we compare images using time slices instead of time in ns, comparing 50000 events, I get 8 events with differences (mostly 1 or 3 pixels different), so 0.016%. Is this acceptable @jsitarek? Also, I made image comparison faster, 50000 events compared in less than 10minutes.

thanks for the test. It looks fine. Let's set the automatic test at the level of 0.03%

ellijobst:
Hi Julian,

in the last few weeks, Alessio and me ran some test_image_comparison tests on these files:
`"20210314_M1_05095172.001_Y_CrabNebula-W0.40+035.root", 
"20210314_M1_05095172.002_Y_CrabNebula-W0.40+035.root",
"20210314_M2_05095172.001_Y_CrabNebula-W0.40+035.root",
"20210314_M2_05095172.002_Y_CrabNebula-W0.40+035.root"`

We ran four different kinds of tests:

1)
for the first one the data was converted into ns in ctapipe_io_magic, and for the cleaning the thresholds for ns were applied. This comparison lead to the following results:

image charge errors: 
0.57%, 0.62%, 0.14% and 0.16% for the four files
image time  errors:
6.71%, 10.89%, 0.87% and 1.12%

All tests failed!

2)
Then we did the same but timeslices were used throughout the entire process, so all thresholds were in timeslices and all files as well. All tests passed with error percentages that were of the order 1e-05.

3)
Then we converted to ns in ctapipe_io_magic, in image_comparison.py converted back to timeslices, then did the cleaning with timeslice thresholds and finally after the cleaning converted back to ns for the comparison.

This gave us:
image charge errors:
0.16%, 0.16%, 0.029% and 0.0368%
image time errors:
0.022%, 0.059%, 0.007% and 0.0%

So for the third file the test passed, since it is below our threshold of 0.03% errors.

4) 
Just to be sure, we tried the 2nd test again, but converted to ns after the cleaning. This gave the same, good, results as before, just as expected.


So it is probably best to just do the cleaning and everything in timeslices and convert to ns afterwards (test2).

What do you think?

jsitarek:
This PR is pretty ancient: it a bunch of  conflicts with the current code, and the tests were already introduced by @Elisa-Visentin some time ago.
Shall we close it? I am not sure if @ellijobst is still available, @aleberti can you also comment?

PR #36, opened 2022-03-28
Add clipping of image

aleberti:
Add image clipping with default value at 750 phe.


YoshikiOhtani:
Hi @aleberti, thanks for spotting on this, and the default value 750 phe is the one used in MARS analysis, is it correct?

aleberti:
actually @ellijobst spotted some images with some pixels with all equal values and then I remembered about the clipping :) and yes, 750 is the MARS default value.

PR #37, opened 2022-03-28
Adding muon ring analysis

gabemery:
Implements a muon ring analysis similar to what is used in lstchain for data. Works on LST and MAGIC MC. Main function should also work for real data (not tested) as in lstchain.

Contains temporary copied code from lstchain master.

example muon ring obtained :
image
image

### DIFF
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_mc_dl0_to_dl1.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_mc_dl0_to_dl1.py
@@ -3,6 +3,7 @@
 
 """
 Author: Yoshiki Ohtani (ICRR, ohtani@icrr.u-tokyo.ac.jp)
+        Muon analysis by Gabriel Emery (gabriel.emery@unige.ch)
 
 This script processes simtel MC DL0 data (*.simtel.gz) containing LST-1 and MAGIC events
 and computes the DL1 parameters (i.e., Hillas, timing and leakage parameters).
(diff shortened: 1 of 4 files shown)


magicctapipe/scripts/lst1_magic/lst1_magic_mc_dl0_to_dl1.py
```diff
@@ -420,6 +475,15 @@ def mc_dl0_to_dl1(input_file, output_dir, config):
     logger.info('\nOutput file:')
     logger.info(output_file)
 
+    if muons_analysis:
+        dir, name = os.path.split(output_file)
+        name = name.replace('dl1', 'muons').replace('LST-1.1', 'LST-1')
```
YoshikiOhtani, line 480: It seems this is a copy from lstchain, but the second replacement assumes that the input is a LST-1 real data file. It would be good to remove it to avoid a confusion.

YoshikiOhtani:
Thank you very much Gabriel for your contribution, and it looks fine to me. There are a few minor comments but anyway it should work without modifying them. It would be great if you could make another pull request once a new version of lstchain is released so that some modules are directly imported from it.

gabemery:
Thanks @YoshikiOhtani , I did the modifications. And the plan is indeed to do another pull request as soon as the upgrade to a new release of lstchain is possible.

YoshikiOhtani:
OK thanks for the commit! Just in case let's ask in tomorrow's meeting if someone wants to comment, otherwise I can merge this to the master.

PR #39, opened 2022-03-31
Fix magic calibration

YoshikiOhtani:
This pull request changes the extractor type from the LocalPeakWindowSum to SlidingWindowMaxSum which is used in MARS analyses, and disables the integration correction. Also, a correction factor is applied to DL1 images which is derived from the difference of the conversion factors 1.165 (= 0.0155/0.0133) between MARS and simtel.

### DIFF
--- a/magicctapipe/scripts/lst1_magic/config.yaml
+++ b/magicctapipe/scripts/lst1_magic/config.yaml
@@ -44,9 +44,13 @@ LST:
 
 MAGIC:
     image_extractor:
-        type: 'LocalPeakWindowSum'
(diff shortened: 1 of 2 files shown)


jsitarek:
thx for implementing it @YoshikiOhtani 
for the moment we can keep it like this, but it might be needed in the next steps to apply somewhat different factors for M1 and M2

YoshikiOhtani:
Hi @jsitarek, yes sure, so far I will recheck the energy correlation by using this factor for both M1 and M2, but let's discuss later if they need to be changed

PR #43, opened 2022-04-20
fixed the script so it works with the current version

jsitarek:
a few of the variable names had to be updated, and some of the imports are now from different files

@aleberti - I'm not sure if this script is still useful, if not we should simply remove it. 
nevertheless I saw similar problems in other files. Also the config files need to be updated, but I was not sure which one is used by which script, hence I did not modify them in this PR


aleberti:
I updated the configuration file. Also, I use unsuitable pixels as calculated in ctapipe-io-magic.

aleberti:
For the moment I merge the PR, and make another one cleaning up the scripts. I will just keep magic_calibrated_to_dl1, since it is a good reference to write out the events in the standard Dl1 format defined in ctapipe.

PR #47, opened 2022-04-26
Add missing console scripts for MAGIC-LST1 analysis

aleberti:
This PR fixes some imports, and adds missing console scripts for MAGIC-LST1 analysis.


YoshikiOhtani (approved):
Thank you @aleberti, I pushes two commits - one to correct the script name in the setup.py and the other to import only the functions that are used in the main functions of the scripts. If you are fine with these changes, please merge this pull request, thanks!

Issue #19, opened 2022-02-04
MAGIC muon analysis with sim_telarray 


jsitarek:
We need to prepare analysis script and do the actual study for muon reflectivity and PSF matching of sim_telarray simulations of MAGIC telescopes

gabemery:
I will contribute on this issue

aleberti:
I see that in PR #37 the muon analysis was added by Gabriel. Should we close this issue? Or should it be closed when scripts for the reflectivity from muon analysis are developed @jsitarek ?

gabemery:
There will be at least one more PR on this issue (upgrade to lstchain 0.9.5+ and script to also analyse real data), and maybe another one with good parameters for the muon analysis function in the config file.

aleberti:
> There will be at least one more PR on this issue (upgrade to lstchain 0.9.5+ and script to also analyse real data), and maybe another one with good parameters for the muon analysis function in the config file.

Thanks! Then let's keep it. I add a task list to keep track:

- [ ] upgrade to lstchain 0.9.5+
- [ ] analyze muons in real data
- [ ] tune parameters for muon analysis

PR #50, opened 2022-04-27
Add muon analysis on calibrated data

gabemery:
Modified the analysis function to work in this case and added a dedicated script.

Related to #19

### DIFF
--- a/magicctapipe/scripts/lst1_magic/config.yaml
+++ b/magicctapipe/scripts/lst1_magic/config.yaml
@@ -64,8 +64,8 @@ MAGIC:
 
     muon_ring:
         thr_low: 25
-        tailcut: [8, 4]
-        ring_completeness_threshold: 15
+        tailcut: [12, 8]
+        ring_completeness_threshold: 25
 
 
 event_coincidence:
--- a/magicctapipe/scripts/lst1_magic/muon_analysis_LST_or_MAGIC_data.py
+++ b/magicctapipe/scripts/lst1_magic/muon_analysis_LST_or_MAGIC_data.py
@@ -20,7 +20,7 @@ import numpy as np
 from pathlib import Path
 
 from astropy.table import Table
-from ctapipe_io_magic import MAGICEventSource
+from ctapipe.io import EventSource
 from lstchain.image.muon import create_muon_table
 from magicctapipe.image import MAGICClean
 from magicctapipe.image.muons import perform_muon_analysis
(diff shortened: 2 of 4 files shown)


magicctapipe/scripts/lst1_magic/config.yaml
```diff
@@ -64,8 +64,8 @@ MAGIC:
 
     muon_ring:
         thr_low: 25
-        tailcut: [8, 4]
```
jsitarek, line 67: was the change of the thresholds intended? 12,8 looks quite large, did it come from your optimization?

magicctapipe/scripts/lst1_magic/muon_analysis_LST_or_MAGIC_data.py
```diff
@@ -0,0 +1,169 @@
...
+    Parameters
+    ----------
+    input_file:
+    output_dir:
+    config:
+    process_run:
+
+    """
+
+    event_source = MAGICEventSource(input_url=input_file)
```
jsitarek, line 52: I think it should work fine if you change to general EventSource, when the file is a MAGIC _Y_ file it will automatically select MAGICEventSource, and this will make it easier to run the same script also on LST

magicctapipe/scripts/lst1_magic/muon_analysis_LST_or_MAGIC_data.py
```diff
@@ -0,0 +1,169 @@
...
+        signal_pixels, image, peak_time = magic_clean.clean_image(
+            event.dl1.tel[tel_id].image, event.dl1.tel[tel_id].peak_time, unsuitable_mask,
+        )
+        perform_muon_analysis(muon_parameters,
+                              event=event,
+                              telescope_id=tel_id,
+                              telescope_name=tel_name,
+                              image=image,
+                              subarray=subarray,
+                              r1_dl1_calibrator_for_muon_rings=
```
jsitarek, line 124: I think putting here directly None except of a separate variable set to None would be more clear

jsitarek (commented):
Thanks @gabemery 
I left a few comments, but they are rather minor. 
More  tricky might be the quality cuts, we use some more in MARS, and here it is only the ring completeness. 
One question, what kind of events were giving you TypeError?

magicctapipe/scripts/lst1_magic/config.yaml
gabemery: It is not optimised, but it followed a comment (by Abelardo?) that if we take only very bright pixels the biases from NSB should be reduced even if the charge extractor is not optimal.

gabemery:
I applied the proposed changes.

PR #52, opened 2022-04-27
Implement the dead time calculation

### LINKED ISSUE #33: Effective time for the software coincidence
While developing codes for creating DL3 files, I noticed how to compute the effective time for the software coincidence. DAQs are performed independently in each telescope system, and the dead time is also different between them. At the moment I use the elapsed time for computing SED. Any ideas are very welcome.

YoshikiOhtani:
This pull request includes the dead time calculation in the combined analysis pipeline. It needs the newest version of ctapipe_io_magic to get the time differences of MAGIC events. It closes #33.

### DIFF
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
@@ -136,23 +136,28 @@ def load_dl2_data_file(input_file, quality_cuts, irf_type):
     n_events = len(df_events.groupby(['obs_id', 'event_id']).size())
     logger.info(f'--> {n_events} stereo events')
 
-    # Calculate the dead time correction factor:
+    # Calculate the dead time correction factor.
+    # For MAGIC we select one telescope which has more number of events than the other:
     logger.info('\nCalculating the dead time correction factor...')
 
+    deadc = 1
     condition = '(time_diff > 0) & (time_diff < 0.1)'
 
     time_diffs_lst = df_events.query(f'(tel_id == 1) & {condition}')['time_diff'].to_numpy()
-    time_diffs_magic = df_events.query(f'(tel_id == 2) & {condition}')['time_diff'].to_numpy()
-
-    deadc = 1
 
     if len(time_diffs_lst) > 0:
         deadc_lst = calculate_deadc(time_diffs_lst, dead_time_lst)
         deadc *= deadc_lst
 
-    if len(time_diffs_magic) > 0:
-        deadc_magic = calculate_deadc(time_diffs_magic, dead_time_magic)
-        deadc *= deadc_magic
+    time_diffs_m1 = df_events.query(f'(tel_id == 2) & {condition}')['time_diff'].to_numpy()
+    time_diffs_m2 = df_events.query(f'(tel_id == 3) & {condition}')['time_diff'].to_numpy()
+
+    if len(time_diffs_m1) >= len(time_diffs_m2):
+        deadc_magic = calculate_deadc(time_diffs_m1, dead_time_magic)
+    else:
+        deadc_magic = calculate_deadc(time_diffs_m2, dead_time_magic)
+
+    deadc *= deadc_magic
 
     dead_time_fraction = 100 * (1 - deadc)
     logger.info(f'--> Dead time fraction: {dead_time_fraction:.2f}%')
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
@@ -44,15 +44,41 @@ ORM_LAT = 28.76177   # unit: [deg]
 ORM_LON = -17.89064   # unit: [deg]
 ORM_HEIGHT = 2199.835   # unit: [m]
 
+dead_time_lst = 7.6e-6   # unit: [sec]
+dead_time_magic = 26e-6   # unit: [sec]
+
 MJDREF = Time(0, format='unix', scale='utc').mjd
 
 __all__ = [
+    'calculate_deadc',
     'load_dl2_data_file',
     'create_event_list',
     'create_gti_table',
     'create_pointing_table',
     'dl2_to_dl3',
 ]
+
+def calculate_deadc(time_diffs, dead_time):
+    """
+    Calculates the dead time correction factor.
+
+    Parameters
+    ----------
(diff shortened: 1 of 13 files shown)


magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
```diff
@@ -108,6 +136,27 @@ def load_dl2_data_file(input_file, quality_cuts, irf_type):
     n_events = len(df_events.groupby(['obs_id', 'event_id']).size())
     logger.info(f'--> {n_events} stereo events')
 
+    # Calculate the dead time correction factor:
+    logger.info('\nCalculating the dead time correction factor...')
+
+    condition = '(time_diff > 0) & (time_diff < 0.1)'
+
+    time_diffs_lst = df_events.query(f'(tel_id == 1) & {condition}')['time_diff'].to_numpy()
+    time_diffs_magic = df_events.query(f'(tel_id == 2) & {condition}')['time_diff'].to_numpy()
```
jsitarek, line 145: M1 and M2 dead time is strongly correlated so it is fine to take only one as you do, but maybe we can make it more "democratic", such that it would work also in the case we have observations when M1 was not working at all.
How about counting the M1 and M2 events and using for this one the one with more events?

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
```diff
@@ -159,9 +206,8 @@ def create_event_list(event_table, effective_time, elapsed_time,
     time_start = Time(event_table['timestamp'][0], format='unix', scale='utc')
     time_end = Time(event_table['timestamp'][-1], format='unix', scale='utc')
 
-    delta_time = time_end.value - time_start.value
-
-    deadc = effective_time / elapsed_time
+    elapsed_time = time_end.value - time_start.value
```
jsitarek, line 209: not sure over how big chunks of data this function can be run, but a safer way to compute the elapsed time is to take all the time difference, exclude big gaps (e.g. > 0.1 s), and sum them

magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py
```diff
@@ -145,8 +146,6 @@ def magic_calib_to_dl1(input_file, output_dir, config, process_run=False):
             logger.warning('\nHot pixels do not exist in a simulation. Setting the "find_hotpixels" option to False...')
             config_cleaning.update({'find_hotpixels': False})
 
-        unsuitable_mask = None
```
jsitarek, line 148: is this change intended?
it looks not related to the dead time

magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py
```diff
@@ -156,13 +155,6 @@ def magic_calib_to_dl1(input_file, output_dir, config, process_run=False):
         )
 
     else:
-        pedestal_type = config_cleaning.pop('pedestal_type')
-
-        if pedestal_type not in pedestal_types:
-            raise KeyError(f'Unknown pedestal type "{pedestal_type}".')
-
-        i_ped_type = np.where(np.array(pedestal_types) == pedestal_type)[0][0]
```
jsitarek, line 164: same here, looks like a merge of different changes

magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py
```diff
@@ -172,9 +164,25 @@ def magic_calib_to_dl1(input_file, output_dir, config, process_run=False):
...
+
+        pedestal_type = config_cleaning.pop('pedestal_type')
+
+        if pedestal_type not in pedestal_types:
+            raise KeyError(f'Unknown pedestal type "{pedestal_type}".')
+
+        i_ped_type = np.where(np.array(pedestal_types) == pedestal_type)[0][0]
+
+    else:
+        unsuitable_mask = None
```
jsitarek, line 184: same here - different change

jsitarek (commented):
sorry for a late feedback. The deadtime calculation looks fine, but I put two sugestions to improve it further (to save us a headache in the future)
there seems to be also some other changes related to hot pixel finding and pedestals in the same PR, I guess those were not intended.

magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py
YoshikiOhtani: Yes indeed it is not related. I recovered back to the original one.

YoshikiOhtani:
Hi @jsitarek, thank you for your checks and comments! I removed the unrelated changes and implemented your suggestions. Regarding the elapsed time, I saw that the cut of the time diffs (< 0.1 sec) is too small for the LST-1 + MAGIC DL3 events. Please find the attached plot of the time difference distribution of DL3 events surviving gammaness cuts (dynamic, g/h efficiency = 0.9). It seems to me that the number of events reconstructed by the three telescopes is actually small, and in addition to this the superior gamma/hadron separation removes background events much better, and so gamma-ray events may already be dominant in the DL3 event list, which makes the time diffs bigger (than LST-1 mono and MAGIC-stereo). So at the moment I don't apply any time diff cuts when calculating the elapsed time, but please let me know what you think.

time_diffs

YoshikiOhtani:
@jsitarek by the way it would be great if you could give an approval when you think it is ok to merge, thanks!

jsitarek (approved):
sorry for late feedback. I approve it if you want to merge it, but I have two remarks:
1) concerning the calculation of the elapsed time - I see your point, this is actually very tricky to get properly if the rate of events is low. But I think it is unlikely that the rate will be dominated by gamma-rays: this would be the case around the source position, but not for the whole camera. If you keep all the events in one basket it should be fine, but if you start dividing into 4 types (following the RF) this indeed can be an issue. The current code with just adding the time differences look dangerous to me - it is quite likely to have some subruns gets missing in  processing and then you have runs with holes in them what are not reflected in the effective time. We could use a larger time difference (10s maybe), or apply a fit and get the elapsed time from that fit, but those would require more tests, so it is fine for me to merge it as it is.
2) github reports that the last line is missing the newline symbol, maybe you can check if it was not removed by mistake

YoshikiOhtani:
Thanks @jsitarek, I added a new line at the end of the file. Regarding the elapsed time, you are right and I overstated that gamma-rays are dominant in a DL3 file. Let's make another pull request when the current code of the elapsed time calculation makes a problem. I will merge this to the master now.

PR #55, opened 2022-04-29
Add output directory to merge_hdf_files.py

aleberti:
Just adds an optional argument to specify the output directory.


YoshikiOhtani (approved):
Thanks @aleberti, this change is of course fine for me, but I would prefer to use None instead of "" for the default output_dir value. I pushed a few commits and so if you don't mind these changes, please merge to the master!

PR #58, opened 2022-05-02
Specify requirement for ipython and ipykernel

### LINKED ISSUE #42: traitlets and jupyter inconsistency?
I got a issue when updating the environment of mcp. 
conda complained on inconsistent traitlets version, i.e. setup.py of MCP requires 'traitlets~=5.0.5', while ipykernel wants > 5.1.
The effect is that while the MCP scripts work, jupyter kernel crashes. 
Anybody had a similar problem? 
Do we really need traitlets in this version, or can it be updated to a newer one?

aleberti:
Specification for ipython and ipykernel versions. Closes #42.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -16,7 +16,7 @@ dependencies:
   - graphviz
   - h5py
   - iminuit>=2
-  - ipython
+  - ipython~=7.31
   - ipywidgets
   - joblib
   - jupyter
--- a/environment.yml
+++ b/environment.yml
@@ -16,7 +16,7 @@ dependencies:
   - graphviz
   - h5py
   - iminuit>=2
-  - ipython~=7.31
+  - ipython
   - ipywidgets
   - joblib
   - jupyter
--- a/setup.py
+++ b/setup.py
@@ -42,6 +42,7 @@ setup(
         'gammapy~=0.19.0',
         'uproot~=4.1',
         'h5py',
+        'ipykernel~=6.4',
         'joblib',
         'matplotlib>=3.5',
         'numba',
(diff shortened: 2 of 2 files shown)


aleberti:
@jsitarek you may try setting up the environment starting from this branch, and let me know if the right versions of ipykernel is installed.

jsitarek:
Hi @aleberti sorry for late feedback on this.
The change that you did still does not work in my case.
It find the requirement, but still ends up with:
ERROR: ipykernel 6.9.1 has requirement traitlets<6.0,>=5.1.0, but you'll have traitlets 5.0.5 which is incompatible.

I can force it if I change ~=6.4 to ==6.4, then it compiles correctly and gives me:
```

In [5]: ipykernel.__version__                                                                                                                                                                                                                                                                                            
Out[5]: '6.4.0'

In [7]: traitlets.__version__                                                                                                                                                                                                                                                                                            
Out[7]: '5.0.5'
```
I think the problem is that ipykernel is installed during the environment creation, and then it goes for version 6.9, making it ~=6.4 means actually >=6.4, ==6.*, so 6.9 is still accepted. Then this 6.9 version is already not compatible with traitlets 5.0.5
So one possible solution is to fix it to ==6.4, and the second one is to make it compatible with ~=6.4.0 (which ends up with 6.4.2 installed). I made a small commit with the second solution

YoshikiOhtani:
Hi, as far as I understand this problem can be solved if we require only the version of ipykernel to 6.4.*, so I pushed a commit to undo the requirement of ipython in the environment yaml file. I tested creating the environment with this settings and it worked without any problems, also for running a script and launching a jupyter kernel. So let me merge this to the master branch and if you face another problem please open a new issue.

PR #59, opened 2022-05-02
modified dl0_to_dl1 script to allow processing files with file names following LSTProd2 style

jsitarek:
the regular expression expected run number then _ and arbitrary text followed with .simtel.gz
but LSTProd2 files have e.g. run999.simtel.gz.
Now the regular expression wants either . or _ after the run number, arbitrary (possibly empty) text and simtel.gz

I checked that it works on the lstprod "train" files, and on the original @YoshikiOhtani proton files (the gamma files of @YoshikiOhtani are using a separate regex)


YoshikiOhtani (approved):
Thanks for the change, I will merge this now.

PR #62, opened 2022-05-05
changed the time limits of the MAGIC cleaning to ns units

jsitarek:
this needs to be correlated with the update of ctapipe_io_magic:
https://github.com/cta-observatory/ctapipe_io_magic/pull/48

### DIFF
--- a/magicctapipe/scripts/lst1_magic/config.yaml
+++ b/magicctapipe/scripts/lst1_magic/config.yaml
@@ -55,8 +55,8 @@ MAGIC:
         use_sum: true
         picture_thresh: 6
         boundary_thresh: 3.5
-        max_time_off: 7.38
-        max_time_diff: 2.46
+        max_time_off: 4.5
+        max_time_diff: 1.5
         find_hotpixels: 'auto'
         pedestal_type: 'from_extractor_rndm'   # select 'fundamental', 'from_extractor' or 'from_extractor_rndm'
 


aleberti (changes requested):
I think also `image/cleaning.py` needs an update, e.g. https://github.com/cta-observatory/magic-cta-pipe/blob/master/magicctapipe/image/cleaning.py#L57 Window2NN (and also 3NN and 4NN) are given in units of time slices (looking at star rc file), and probably also line https://github.com/cta-observatory/magic-cta-pipe/blob/master/magicctapipe/image/cleaning.py#L586

aleberti (changes requested):
Please also change the version number of ctapipe_io_magic in setup.py and environmen.yml, as I created the v0.4.4 release yesterday.

jsitarek:
thank you @aleberti, you are perfectly right, I forgot about those time windows. This probably explains why I saw some small differences. I will update this and recheck it. Now that you did the release I will update the ctapipe_io_version as well. Concerning the condition in teh interpolating function, this looks like a very broad condition (much larger than the whole window size), and I'm not sure why it is there (in the interpolate_fast function it is not), I think we can keep it as it is (or revise the whole function).

jsitarek:
updated those window times as well, 
I couldn't read the sampling speed in the cleaning class, because it does not deal with the files directly, and adding an extra member to pass this information seems excessive since it is only for updating the default values that are hardcoded, so I hardcode that sampling speed there as well. 
Now the agreement of image parameters (I check intensity and width) on the test file is perfect between the old and new version and only the time slope is recomputed. 
@aleberti please have a look and let's merge it if you think it is fine as it is

aleberti:
> thank you @aleberti, you are perfectly right, I forgot about those time windows. This probably explains why I saw some small differences. I will update this and recheck it. Now that you did the release I will update the ctapipe_io_version as well. Concerning the condition in teh interpolating function, this looks like a very broad condition (much larger than the whole window size), and I'm not sure why it is there (in the interpolate_fast function it is not), I think we can keep it as it is (or revise the whole function). 

About the time interpolation, the broad condition is taken from the corresponding function in MBadPixelTreat.cc in MARS. As far as I know the interpolate_fast function is not used even if implemented

aleberti (approved):

PR #67, opened 2022-05-23
fix for training with LSTProd2

jsitarek:
in LSTProd2 some of the pointing nodes are generated with the same obs_id.
This causes confusion in RF generation script because it is using event_id and obs_id to identify the events,
and thus you can mix up events from different pointings as parts of one event
Now grouping of events is done in addition also by true values of alt and az

matching by pointing alt/az would be probably better, but those are the values that we want to use in the training and using the same column as grouping index and column label is problematic

@YoshikiOhtani , can you have a look at this PR?

### DIFF
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_train_rfs.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_train_rfs.py
@@ -52,6 +52,10 @@ tel_combinations = {
 event_class_gamma = 0
 event_class_bkg = 1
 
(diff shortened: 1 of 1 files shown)


YoshikiOhtani:
Hi @jsitarek, I think this change should work for the old MCs so there will be no problem, but setting non-int values for the pandas index would be not ideal. Just let me confirm, aren't there any IDs labeled to each pointing node? If not, I will approve this pull request. Thanks!

Issue #69, opened 2022-05-27
Averaging telescope-wise parameters


YoshikiOhtani:
The current combined analysis pipeline averages telescope-wise parameters before creating the IRFs and DL3 files, for example at the following line:
https://github.com/cta-observatory/magic-cta-pipe/blob/d05e57102098c4bce808f0e0c96db5a33bc6b782/magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py#L133

This uses the function `get_dl2_mean` which averages the DL2 parameters with the inverse of the uncertainties of the DL2 parameters:
https://github.com/cta-observatory/magic-cta-pipe/blob/d05e57102098c4bce808f0e0c96db5a33bc6b782/magicctapipe/utils/functions.py#L333-L334

We may want to optimize the averaging method, especially what weights would make the performance best. For example, we can try to use the "intensity" parameter for weighing brighter images, but it in general puts weights to LST-1 images. Or we may also try to use different gammaness cuts to each telescope, not averaging them, for example g > 0.8 for LST-1 and g > 0.7 for MAGICs.

Any comments or suggestions are very welcome!

jsitarek:
as we discussed via e-mail @YoshikiOhtani one of the other things to test would be the standard statistical weighting with the inverted variance (1/error^2) instead of 1/error

YoshikiOhtani:
By the way the current pipeline defines the error (or uncertainty) of a RF regressor output as the standard deviation of whole decision trees outputs. Please find the attached plot for an example of the energy estimation. The RF output is actually consistent with the mean of the trees outputs. Since there seems to be no method in the scikit learn module (sklearn.ensemble.RandomForestRegressor) to estimate the uncertainty, the std is computed within our script. I think it would be fine to use the std for the error, but please let me know if you have any doubts about this.

![energy_regressor](https://user-images.githubusercontent.com/39613098/170882596-a9f676b7-f68a-40bc-8822-a9027dcf0546.png)

jsitarek:
taking just STD of the distribution from the trees is not a bad measure for this purpose, but I think the actual value will be strongly overestimated (you can also see it from the plot above, the difference of the true and estimated value is much smaller than this std), because you do an average over many trees. The error of the average (which would be std / sqrt(Ntrees) would be likely closer to the true (but most likely it will be underestimating the error), but for the  purpose of the averages this does not make any difference since it is just a constant factor that would cancel out

[...]

PR #70, opened 2022-05-27
Fixed a bug of using non-inverse error for weights

YoshikiOhtani:
When I was looking into the issue #69, I found that when averaging the telescope-wise reconstructed directions the pipeline uses the errors which are not inverse. This makes the reconstruction performance worse because it weights to the reco parameter whose error is large. I checked that fixing this issue and using the inverse errors improve the angular resolutions, please see the attached two plots for the comparison of the resolutions before and after fixing the bug.

angres_magic_bugfix
angres_mlst_bugfix

@jsitarek, since it should affect your analysis, please give me an approval when you acknowledge this fix. Thanks!

### DIFF
--- a/magicctapipe/utils/functions.py
+++ b/magicctapipe/utils/functions.py
@@ -73,7 +73,7 @@ def calculate_mean_direction(lon, lat, weights=None):
     lon: pandas.core.series.Series
         Longitude in a spherical coordinate
     lat: pandas.core.series.Series
-        Latitude in a spherical coodinate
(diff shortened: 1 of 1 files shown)


jsitarek (changes requested):
the fix looks fine, but I think the same error is also below in the code in line 376 when you average the RA/DEC position

YoshikiOhtani:
Thanks @jsitarek for your check, I overlooked it. The correction is done!

jsitarek (approved):

PR #71, opened 2022-05-31
Change the definition of the DL2 parameter uncertainty

YoshikiOhtani:
It is still a draft pull request related to the issue #69.

### DIFF
--- a/magicctapipe/reco/estimators.py
+++ b/magicctapipe/reco/estimators.py
@@ -461,11 +461,9 @@ class EventClassifier:
             x_predict = df_events[self.features].to_numpy()
             gammaness = self.telescope_rfs[tel_id].predict_proba(x_predict)[:, 0]
 
-            responces_per_estimator = []
-            for estimator in self.telescope_rfs[tel_id].estimators_:
-                responces_per_estimator.append(estimator.predict_proba(x_predict)[:, 0])
-
-            gammaness_uncert = np.var(responces_per_estimator, axis=0)
+            gammaness_uncert = gammaness * (1 - gammaness)
+            gammaness_uncert[gammaness == 1] = 0.99 * (1 - 0.99)
+            gammaness_uncert[gammaness == 0] = 0.01 * (1 - 0.01)
 
             df_reco_class = pd.DataFrame(
                 data={'gammaness': gammaness,
--- a/magicctapipe/reco/estimators.py
+++ b/magicctapipe/reco/estimators.py
@@ -106,11 +106,11 @@ class EnergyRegressor:
             for estimator in self.telescope_rfs[tel_id].estimators_:
                 responces_per_estimator.append(estimator.predict(x_predict))
 
-            reco_energy_err = np.var(responces_per_estimator, axis=0)
+            reco_energy_uncert = np.var(responces_per_estimator, axis=0)
 
             df_reco_energy = pd.DataFrame(
                 data={'reco_energy': reco_energy,
-                      'reco_energy_err': reco_energy_err},
+                      'reco_energy_uncert': reco_energy_uncert},
                 index=df_events.index,
             )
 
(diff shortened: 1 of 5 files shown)


YoshikiOhtani:
Sorry for pushing again some commits, but I have also updated the scripts of creating the IRFs and DL3 files so that one can select the type of DL2 weight. Also, I changed the default weight to None (means weight = 1) for a safe side.

YoshikiOhtani:
Sorry for pushing again some commits, but I have also updated the scripts of creating the IRFs and DL3 files so that one can select the type of DL2 weight. Also, I changed the default weight to None (means weight = 1) for a safe side.

magicctapipe/reco/estimators.py
```diff
@@ -459,10 +459,15 @@ def predict(self, event_data):
...
+            gammaness = self.telescope_rfs[tel_id].predict_proba(x_predict)[:, 0]
+
+            gammaness_uncert = gammaness * (1 - gammaness)
+            gammaness_uncert[gammaness == 1] = 0.99 * (1 - 0.99)
+            gammaness_uncert[gammaness == 0] = 0.01 * (1 - 0.01)
 
             df_reco_class = pd.DataFrame(
-                data={'gammaness': responses[:, 0]},
+                data={'gammaness': gammaness,
+                      'gammaness_uncert': gammaness_uncert},
```
jsitarek, line 470: uncert==>var

magicctapipe/scripts/lst1_magic/config.yaml
```diff
@@ -181,6 +181,7 @@ event_classifier:
 create_irf:
     quality_cuts: null
     irf_type: 'software'   # select 'hardware', 'software', 'software_with_any2' or 'magic_stereo'
+    dl2_weight: null   # type of the weight for averaging tel-wise DL2 parameters, null, 'uncert' or 'intensity'
```
jsitarek, line 184: maybe also here uncert ==> var

magicctapipe/utils/functions.py
```diff
@@ -324,25 +325,48 @@ def get_dl2_mean(event_data):
...
     is_simulation = ('true_energy' in event_data.columns)
+
     group_mean = event_data.groupby(['obs_id', 'event_id']).mean()
 
+    if weight == None:
+        gammaness_weights = pd.Series(np.repeat(1, len(event_data)), index=event_data.index)
+        energy_weights = pd.Series(np.repeat(1, len(event_data)), index=event_data.index)
+        direction_weights = pd.Series(np.repeat(1, len(event_data)), index=event_data.index)
+
+    elif weight == 'uncert':
```
jsitarek, line 336: uncert ==> var

magicctapipe/utils/functions.py
```diff
@@ -373,7 +397,7 @@ def get_dl2_mean(event_data):
         reco_ra_mean, reco_dec_mean = calculate_mean_direction(
```
jsitarek, line 397: even while your current change is just reproducing the old behaviour, I have some doubts concerning this part. 
averaging is done first for az/alt direction and here you are doing it again for lon/lat coordinates. 
It is more or less fine, because for the camera size the spherical plane is nearly flat and thus a rotation of the two coordinate systems would end up as a linear transformation, but the spherical shape I think would still produce some small differences (I guess mostly pronounced for large lat) that would end up that reconstructed az/alt direction and lat/lon direction slightly mismatch one another. 
I think it should be simple to just use the coordinate transformation to go from averaged alt/az to lat/lon. This would also make it easier to maintain the function later on. 
As you prefer you might change it immediately here, or keep it as it is and we rediscuss it in a separate issue.

jsitarek (commented):
Hi @YoshikiOhtani 
Thanks for the change.
I suggested to rename one of the variables 
And I also have a side comment concerning the lat/lon averaging.Other than this it is fine with it.

YoshikiOhtani:
Hi @jsitarek, thank you very much for your reviews, I pushed the change of the variable names and also the transformation of the mean Alt/Az to the RA/Dec coordinate instead of computing the mean RA/Dec coordinate. Would be great if you could give an approval in case the changes are fine for you. Thanks!

YoshikiOhtani:
I'm sorry but I mistakenly/accidentally merged this branch to the master... please let me know if you find some problems...!

PR #77, opened 2022-06-25
Refactor stereo reconstruction

YoshikiOhtani:
This pull request includes some minor refactoring of the code of the stereo parameters reconstruction. Here is a brief summary of the changes:

- changed the option name of the function `stereo_reconstruction` from `magic_only` to `magic_only_analysis` (but the argument name `--magic-only` used when running the script does not change).
- cleaned up the code of checking the angular distance of the pointing directions
- modified the logger information display

I checked that the outputs do not change before and after the refactoring and so it would be safe to merge now, but again I keep it here a few days so any comments or requests are very welcome.

### DIFF
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_stereo_reco.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_stereo_reco.py
@@ -2,15 +2,19 @@
 # coding: utf-8
 
 """
-This script processes DL1 events and reconstructs the stereo parameters with more than one telescope information.
-The quality cuts specified in a configuration file are applied to events before the reconstruction.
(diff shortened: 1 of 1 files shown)


aleberti:
Also here changes look good, no particular comment from my side

YoshikiOhtani:
I just pushed a commit about the change of an output file name when using the `--magic-only` argument to clarify from name that the data are processed by only MAGICs.

YoshikiOhtani:
@aleberti I also modified the stereo reco script by following suggestions by flake8 (max-line-length = 90), and run isort and black. It would be much appreciated if you could check it and merge to the master if it looks OK to you. Thanks!

aleberti (approved):

PR #78, opened 2022-06-29
Refactor training/applying RFs 

YoshikiOhtani:
The main purpose of this pull request was to implement an option to use unsigned features for training/applying RFs, but at the same time I refactored the code a bit. Here is a briefly summary of the changes that I did:

- Added the option "--use-unsigned" with which unsigned features are used for training and applying RFs.
- Drop NaN values from an input data frame inside the estimator modules, instead of doing it in the load functions.
- Use the CameraFrame stored in a subarray instead of creating it inside the DirectionEstimator
- Changed the argument name of lst1_magic_dl1_stereo_to_dl1.py, from `--input-file` to `--input-file-dl1`
- Changed the logger information display

Any comments are very welcome.

### DIFF
--- a/magicctapipe/reco/estimators.py
+++ b/magicctapipe/reco/estimators.py
@@ -60,6 +60,8 @@ class EnergyRegressor:
 
         self.telescope_rfs.clear()
 
+        event_data.dropna(subset=self.features, inplace=True)
+
         if self.use_unsigned_features:
             event_data[self.features] = np.abs(event_data[self.features])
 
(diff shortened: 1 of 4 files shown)


YoshikiOhtani:
These scripts are also refactored by following suggestions by flake8, isort and black, so ready for reviews. @aleberti would be great if you could briefly check them and merge to the master if it is OK to you. Thanks!

aleberti (approved):

PR #81, opened 2022-07-01
Implement IRF interpolation method

jsitarek:
this still does not work too good. I will check if there is no problem with interpolation. Nevertheless @YoshikiOhtani any feedback on the code is very much appreciated.

### DIFF
--- a/magicctapipe/scripts/lst1_magic/config.yaml
+++ b/magicctapipe/scripts/lst1_magic/config.yaml
@@ -199,8 +199,8 @@ create_irf:
 
     energy_bins:   # log space
         start: 0.01   # unit: [TeV]
-        stop: 100   # unit: [TeV]
-        num: 21   # number of edges (nbins = num - 1)
+        stop: 1000   # unit: [TeV]
+        num: 26   # number of edges (nbins = num - 1)
 
     migration_bins:   # log space
         start: 0.2
--- a/magicctapipe/scripts/lst1_magic/config.yaml
+++ b/magicctapipe/scripts/lst1_magic/config.yaml
@@ -217,3 +217,4 @@ dl2_to_dl3:
     source_name: 'Crab'
     source_ra: null   # used when the source name is null
     source_dec: null   # used when the source name is null
+    interpolation_method: 'nearest' # 'nearest' - no interpolation (closest point taken), 'linear', 'cubic' - interpolation
--- a/magicctapipe/scripts/lst1_magic/config.yaml
+++ b/magicctapipe/scripts/lst1_magic/config.yaml
@@ -220,14 +220,14 @@ create_irf:
     gammaness:
         cut_type: 'dynamic'   # select 'global' or 'dynamic'
         global_cut_value: 0.8   # used for the global cut
-        gamma_efficiency: 0.9   # used for the dynamic cut
+        efficiency: 0.9   # used for the dynamic cut
         min_cut: 0.05   # used for the dynamic cut
         max_cut: 0.85   # used for the dynamic cut
 
     theta:
         cut_type: 'global'   # select 'global' or 'dynamic'
         global_cut_value: 0.2   # used for the global cut
-        gamma_efficiency: 0.75   # used for the dynamic cut
+        efficiency: 0.75   # used for the dynamic cut
         min_cut: 0.1   # used for the dynamic cut
         max_cut: 0.45   # used for the dynamic cut
 
--- a/magicctapipe/scripts/lst1_magic/config.yaml
+++ b/magicctapipe/scripts/lst1_magic/config.yaml
@@ -233,7 +233,7 @@ create_irf:
 
 
 dl2_to_dl3:
-    interpolation_method: 'nearest' # 'nearest' - no interpolation (closest point taken), 'linear', 'cubic' - interpolation
+    irf_interpolation_method: 'nearest' # 'nearest' - no interpolation (closest point taken), 'linear', 'cubic' - interpolation
     source_name: 'Crab'
     source_ra: null   # used when the source name is null
     source_dec: null   # used when the source name is null
--- a/magicctapipe/scripts/lst1_magic/config.yaml
+++ b/magicctapipe/scripts/lst1_magic/config.yaml
@@ -185,8 +185,8 @@ create_irf:
 
     energy_bins:
         start: 0.01   # unit: [TeV]
-        stop: 100   # unit: [TeV]
-        n_bins: 20   # number of bins in log space
+        stop: 1000   # unit: [TeV]
+        n_bins: 25   # number of bins in log space
 
     migration_bins:
         start: 0.2
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
@@ -371,7 +371,7 @@ def create_irf(
 
     table_gamma, sim_info_gamma = load_dl2_data_file(input_file_gamma, quality_cuts, irf_type, dl2_weight)
 
-    if sim_info_gamma.viewcone.value != 0.0:
+    if sim_info_gamma.viewcone.value > 1.0:
         logger.info('\nHave not yet implemented functions to create diffuse IRFs. Exiting.')
         sys.exit()
 
@@ -592,7 +592,7 @@ def create_irf(
     # Save the data in an output file:
     Path(output_dir).mkdir(exist_ok=True, parents=True)
 
-    regex = r'dl2_gamma_(\S+)_run.*'
+    regex = r'dl2_(\S+)_run.*'
     file_name = Path(input_file_gamma).name
 
     if re.fullmatch(regex, file_name):
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
@@ -150,7 +150,7 @@ def load_dl2_data_file(
 
     pointing_zd = np.mean(90 - event_table["pointing_alt"].to_value(u.deg))
     pointing_az = np.mean(event_table["pointing_az"].to_value(u.deg))
-    pointing = np.array([np.round(pointing_zd, 3), np.round(pointing_az, 3)])
+    pointing = np.array([pointing_zd.round(3), pointing_az.round(3)])
 
     # Load the simulation configuration:
     sim_config = pd.read_hdf(input_file, key="simulation/config")
@@ -251,7 +251,8 @@ def apply_dynamic_theta_cut(
 ):
     """
     Applies dynamic (energy-dependent) theta cuts to input events.
-    The cuts are computed in each energy bin so as to keep the specified gamma efficiency.
+    The cuts are computed in each energy bin so as to keep the specified
+    gamma efficiency.
 
     Parameters
     ----------
@@ -424,7 +425,7 @@ def create_irf(
         "INSTRUME": "LST-1_MAGIC",
         "FOVALIGN": "RADEC",
         "PNT_ZD": (pnt_gamma[0], "deg"),
-        "PNT_AZ": (pnt_gamma[0], "deg"),
+        "PNT_AZ": (pnt_gamma[1], "deg"),
         "IRF_TYPE": irf_type,
     }
 
@@ -456,7 +457,7 @@ def create_irf(
     elif gam_cut_type == "dynamic":
         logger.info("\nApplying the dynamic gammaness cuts:")
 
-        gh_efficiency = config_irf["gammaness"]["gamma_efficiency"]
+        gh_efficiency = config_irf["gammaness"]["efficiency"]
         gh_cut_min = config_irf["gammaness"]["min_cut"]
         gh_cut_max = config_irf["gammaness"]["max_cut"]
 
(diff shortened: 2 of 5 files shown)


magicctapipe/scripts/lst1_magic/config.yaml
```diff
@@ -185,8 +185,8 @@ create_irf:
 
     energy_bins:
         start: 0.01   # unit: [TeV]
-        stop: 100   # unit: [TeV]
-        n_bins: 20   # number of bins in log space
+        stop: 1000   # unit: [TeV]
```
jsitarek, line 188: I increased it here, because I had crashes with events going outside of energy range, but we might want to have a different type of fix instead - such that the narrower binning than MC range can be used

magicctapipe/scripts/lst1_magic/config.yaml
```diff
@@ -217,3 +217,4 @@ dl2_to_dl3:
     source_name: 'Crab'
     source_ra: null   # used when the source name is null
     source_dec: null   # used when the source name is null
+    interpolation_method: 'nearest' # 'nearest' - no interpolation (closest point taken), 'linear', 'cubic' - interpolation
```
jsitarek, line 220: in the test I got better results with "nearest" than with "linear", which is quite disturbing, because it just takes the closest point without any interpolation

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
```diff
@@ -371,7 +371,7 @@ def create_irf(
 
     table_gamma, sim_info_gamma = load_dl2_data_file(input_file_gamma, quality_cuts, irf_type, dl2_weight)
 
-    if sim_info_gamma.viewcone.value != 0.0:
+    if sim_info_gamma.viewcone.value > 1.0:
```
jsitarek, line 374: "Test" gamma ray MCs are produced with viewcone of 0.4 deg and were failing this check

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
```diff
@@ -592,7 +592,7 @@ def create_irf(
     # Save the data in an output file:
     Path(output_dir).mkdir(exist_ok=True, parents=True)
 
-    regex = r'dl2_gamma_(\S+)_run.*'
+    regex = r'dl2_(\S+)_run.*'
```
jsitarek, line 595: this is also because in the "Test" production files there is no explicit "gamma"

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
```diff
@@ -354,8 +365,152 @@ def create_pointing_table(event_table):
...
+            ghcuts_all.append(ghcuts.data['cut'])
+            
+            ghcuts_low_last=ghcuts_low
+            ghcuts_high_last=ghcuts_high
+            ghcuts_center_last=ghcuts_center
+        else:
+            logger.warning("skipping "+ name)
+
+    # fix me! only the last file is checked, no consistency checks
+    extra_headers={key: aeff.header[key] for key in list(set(extra_headers_list) & set(aeff.header.keys()))}
```
jsitarek, line 471: this I should probably do better to check that indeed the settings are all the same in all the files, but it is somewhat tedious, because one needs to check both if the list of keys is the same, and then only if the values match

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
```diff
@@ -354,8 +365,152 @@ def create_pointing_table(event_table):
...
+    
+    # here we need to swap axes because the function expects shape: (n_grid_points, n_energy_bins, n_migration_bins, n_fov_offset_bins)
+    edisp_interp=interp.interpolate_energy_dispersion(np.swapaxes(edisp_all, 1, 3), points, target, method=method)
+    
+    # now create the HDUS
+    # to have the same format we need to loose one dimention, this might need to be rewised for files with many bins in offset angle!
+    aeff_hdu=create_aeff2d_hdu(aeff_interp[...,0], aeff_energ_bins[0], aeff_theta_bins[0], extname=aeff_ext_name, **extra_headers)
+    edisp_hdu=create_energy_dispersion_hdu(edisp_interp[0,...], edisp_energ_bins[0], edisp_migra_bins[0], edisp_theta_bins[0], extname=edisp_ext_name)
+    
+    ghcuts_interp = griddata(points, ghcuts_all, target, method=method)
```
jsitarek, line 503: there is no function for interpolation of cut values in pyirf yet, so I implemented it directly here for the time being using the same interpolation engine as Aeff and edisp above. 
Note that if we run with variable theta2 cut it would also need to be interpolated (not done yet)

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
```diff
@@ -476,8 +643,13 @@ def main():
...
     parser.add_argument(
-        '--input-file-irf', '-i', dest='input_file_irf', type=str, required=True,
-        help='Path to an input IRF file.',
+        '--input-file-irf', '-i', dest='input_file_irf', type=str, required=False, default=None,
+        help='Path to an input IRF file (single).',
+    )
+
+    parser.add_argument(
+        '--input-dir-irf', dest='input_dir_irf', type=str, required=False,
+        help='Path to an input IRF directory (interpolation will be applied).',
```
jsitarek, line 652: I meant here to keep the original API: depending if file or dir is passed it used a single file or does the interpolation. We could pass the dir with the same variable and then check if it is a file or a dir, but this is a bit confusing in my opinion

magicctapipe/scripts/lst1_magic/config.yaml
YoshikiOhtani: So far I keep this change, but would be good to know what kind of crashes happened

magicctapipe/scripts/lst1_magic/config.yaml
YoshikiOhtani: Thanks, I keep the default configuration as "nearest"

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
YoshikiOhtani: Here I changed the lines so that the script does not get the information from the input file name, but reads the pointing direction from the input data and use it for the output file name

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
YoshikiOhtani: Regarding this I implemented a function to check the consistency of the IRF data and the headers of the input IRF files.

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
YoshikiOhtani: Thanks, I also implemented the interpolation for the theta (= rad_max) cuts.

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
```diff
@@ -422,10 +586,13 @@ def dl2_to_dl3(input_file_dl2, input_file_irf, output_dir, config):
...
 
-    # Create a event list HDU:
+    # Create an event list HDU:
     logger.info('\nCreating an event list HDU...')
 
-    event_list, event_header = create_event_list(event_table, deadc, **config_dl3)
+    source_name=config_dl3['source_name']
+    source_ra=config_dl3['source_ra']
+    source_dec=config_dl3['source_dec']
+    event_list, event_header = create_event_list(event_table, deadc, source_name, source_ra, source_dec)
```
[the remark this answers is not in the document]
YoshikiOhtani: Instead of the event list, I added the method to the headers of the IRF HDUs.

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
YoshikiOhtani: Regarding this I simply removed the argument "--input-file-irf". I checked that even if the input IRF file is one, the script works by selecting the interpolation method to "nearest".

YoshikiOhtani:
Apologies for the late feedback but thank you very much for implementing! I checked the code and in general I agree with the procedure. However, I have an idea of making it more logical and smooth, and since Julian is taking holidays I pushed some commits to the branch instead of only commenting. What I mainly did is to implement the function `load_irf_files` not only to load the input IRF files but also to check the consistency of the configuration, such as the binning of the IRF data and the settings stored in the headers. I also changed the code so that loading of the files, interpolating the IRFs and create the HDUs are performed inside the `dl2_to_dl3` function. The interpolation of the dynamic theta cuts (= rad_max) is also implemented by using the same function `scipy.interpolate.griddata` as the other interpolation methods. The argument "--input-irf-file" is now replaced with "--input-irf-dir", and even if only one IRF file is input the script works by specifying the `irf_interpolation_method` setting to "nearest". 

@aleberti since I also pushed the commits as a developer, could you please have a look at this as a reviewer when you have time? Thanks in advance!

aleberti:
Hi @YoshikiOhtani, sure I can review. Maybe it would be good if you run the new interpolations on a dataset you already have and check the performance (e.g. how it compares to old methods/no interpolation or if you see strange features)? However I do not know if you have data locally, maybe you have everything in the IT container which is currently down.

YoshikiOhtani:
Hi @aleberti, thanks! Exactly, I did not transfer any data to other machines, except only a DL2 MC data file produced with a fixed pointing direction that I have in my local PC. I tested creating the IRFs with the data and the `create_irf` script of this branch and it succeeded. Also, using the interpolation functions with the one IRF file and the "nearest" method worked fine as before implementing the interpolation methods, so I think the functionality of the script is safely protected. But for testing "linear" and "cubic" options more IRF files are needed, which I don't have now...

aleberti:
Regarding this PR, should it be merged now or should we wait for Julian's opinion? Actually I will see him next week at TeVPA (he is in the participant list), so I can ask him what he thinks about it.

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
```diff
@@ -377,39 +395,45 @@ def create_irf(
...
         bkg_fov_offset_bins = np.linspace(**config_irf["bkg_fov_offset_bins"]) * u.deg
 
-    extra_headers = {
+    extra_header = {
         "TELESCOP": "CTA-N",
         "INSTRUME": "LST-1_MAGIC",
         "FOVALIGN": "RADEC",
-        "QUAL_CUT": quality_cuts,
+        "PNT_ZD": (pnt_gamma[0], "deg"),
+        "PNT_AZ": (pnt_gamma[0], "deg"),
```
jsitarek, line 427: I think it should be [1] here instead of [0]

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
```diff
@@ -467,28 +493,30 @@ def create_irf(
...
-        extra_headers["RAD_MAX"] = (global_theta_cut.value, "deg")
+        extra_header["RAD_MAX"] = (global_theta_cut.value, "deg")
 
     elif theta_cut_type == "dynamic":
         logger.info("\nApplying the dynamic theta cuts:")
 
-        gamma_efficiency = config_irf["theta"]["gamma_efficiency"]
-        min_cut = u.Quantity(config_irf["theta"]["min_cut"], u.deg)
-        max_cut = u.Quantity(config_irf["theta"]["max_cut"], u.deg)
+        theta_efficiency = config_irf["theta"]["gamma_efficiency"]
```
jsitarek, line 506: I guess "gamma_efficiency" here should be "theta_efficiency"

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
```diff
@@ -331,137 +518,241 @@ def create_pointing_table(event_table):
...
+    # Load the input DL2 data file:
+    event_table, deadc = load_dl2_data_file(
+        input_file_dl2, quality_cuts, irf_type, dl2_weight
+    )
 
-    # Load the input IRF file:
-    logger.info('\nLoading the input IRF file:')
-    logger.info(input_file_irf)
+    pointing_coszd = np.mean(np.sin(event_table["pointing_alt"]))
+    pointing_az = np.mean(event_table["pointing_az"])
```
jsitarek, line 580: I realized that we might have  a problem here if the azimuth makes a full 2 pi turn. One could check the .std of the azimuth in two definition -180 - 180 and 0-360 and use the smaller one for calculations (then translate it into the default convention of azimuth).  But we can leave it as it is (maybe with a "FIX ME" written in the code), or just an error if .std is greater than  lets say 20 deg

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
```diff
@@ -471,38 +762,54 @@ def main():
...
     )
 
     parser.add_argument(
-        '--input-file-irf', '-i', dest='input_file_irf', type=str, required=True,
-        help='Path to an input IRF file.',
+        "--input-dir-irf",
+        "-i",
+        dest="input_dir_irf",
+        type=str,
+        required=False,
```
jsitarek, line 778: I set required to False before because I used two different options for single file and directory, and only one should be used, but since now the code used only -dir- option, I think it should be again required

jsitarek (commented):
Hi @YoshikiOhtani 
sorry for a late reply.
I went through the updated PR and put a few more comments. Other than this, it is fine with me.

YoshikiOhtani:
Hi @jsitarek, sorry for being late to reply, but thank you very much for your checks. I have just pushed some commits to fix them. Regarding the problem of azimuth 2 pi turn I just added FIX ME for now, but I may push a commit to fix it later. So let's merge this to the master now. If you face some problems please let me know.

PR #83, opened 2022-07-14
New MC muons analysis script + non hard-coded plot creation option

gabemery:


YoshikiOhtani:
OK thanks, now I'm refactoring the script with the pull request #75 so there I will remove the corresponding lines.

PR #85, opened 2022-07-20
Fix the problem of installing the software with a Max OS

YoshikiOhtani:
Installing the software in my local Mac OS failed due to a problem of installing protozfits with pip. It was solved by installing the module with conda before installing lstchain with pip. Also, I cleaned up the environment.yml file a bit as follows:

- Remove the cta-observatory channel and instead add conda-forge channel, and then clean some lines
- Install uproot with conda instead of pip, because it is available from conda-forge

I think this changes do not harm installations for other systems. I wanted to try it just in case for linux, but now the IT container is not accessible..  I assign @aleberti for a reviewer so it would be great if you could have a look at them and merge this if it is OK for you. Thanks!

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -41,6 +41,7 @@ dependencies:
   - sphinx_rtd_theme
   - tqdm
   - traitlets
+  - uproot~=4.1
   - vitables
   - wheel
   - xz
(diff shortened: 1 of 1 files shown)


aleberti:
Installation was succesfull also on Ubuntu 20.04 LTS, so I will merge this.

PR #87, opened 2022-07-25
Refactor combined analysis functions

YoshikiOhtani:
With this pull request the functions used for the combined analysis are refactored by following suggestions of flake8, isort and black. Any reviews are very much appreciated.


aleberti:
> With this pull request the functions used for the combined analysis are refactored by following suggestions of flake8, isort and black. Any reviews are very much appreciated.

No problems for me. I think only thing is that in the diff I see that for functions.py it complains that no new line is present at the end of the file. Can you add it? Apart from that, fine for me!

YoshikiOhtani:
Thanks Alessio, regarding the function.py I checked that there was no new line before refactoring the code but now there is, so it should be no problem.

PR #91, opened 2022-08-19
Random updates and refactoring of the combined analysis pipeline 

YoshikiOhtani:
This pull request includes some random updates and refactoring of the combined analysis pipeline. I tested running the updated pipeline with a few data samples and it worked. However, I will keep this pull request open until I confirm that the pipeline produces the consistent results with the whole Crab samples that I usually analyze. Meanwhile, it would be very much appreciated if you could give me some comments or feedbacks. Let me assign @aleberti and @jsitarek as reviewers.

Here let me summarize the changes as follows:

**General changes**
- Refactored the pipeline by following suggestions of flake8, isort and black with the `max-line-length = 88` setting (72 for docstrings as following PEP8). Constant variables are now defined in capital letters. Please note that the most updated version of black consider the [magic trailing comma](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#the-magic-trailing-comma) by default, but in this pull request I removed this functionality with the option `-C` (or `--skip-magic-trailing-comma`) to reduce the user flexibility as much as possible. 
- Created `magicctapipe/io` directory and the modules `container.py`, `gadf.py` and `io.py`, and moved some functions to them from `magicctapipe/scripts/lst1_magic/*.py` and `magicctapipe/utils/functions.py`.
- Updated the documentation of the scripts and added explanations to the code as much as possible so that users can understand what is done in the pipeline.
- Updated the logger information display so that while running the scripts users can easily cross-check if the settings are correct or not.
- Naming of the output files is simplified, i.e., just replacing specific characters assuming that the input file names do not change from the default ones.

**Important changes for users**
- Changed the format of some settings in the configuration file to string in case they have units, so that astropy `Quantity` can read the unit directly from the string. See the following example:
https://github.com/cta-observatory/magic-cta-pipe/blob/95f311ea3536b2571a5e514eb1a4007c6491d357/magicctapipe/scripts/lst1_magic/config.yaml#L79
- Changed the argument to train RFs from `--train-direction` to `--train-disp`, and also the name of the output RF file from `direction_regressors_*.joblib` to `disp_regressors_*.joblib`.
- Changed the parameter name `dist_sum` to `disp_diff_sum`, which is the sum of the angular distances of DISP candidates.
- Changed the configuration name from `software_with_any2` to `software`, and `software` to `software_only_3tel`, since the analysis with any2 events would be more standard for the software coincidence.
- **(EDIT)** Added the parameter `disp_diff_mean` which is the mean of the angular distances of DISP candidates.
- **(EDIT)** Changed the argument of `lst1_magic_train_rfs.py` from `--input-file-gamma`, `--input-file-bkg` to `--input-dir-gamma`, `--input-dir-proton` so that multiple input MC files produced by different pointing directions can be read.

From here let me describe the other noticeable changes for each script or module:

**functions.py**
- Renamed `calculate_angular_distance` to `calculate_off_coordinates` and simplify the functionality to return only OFF coordinates. Please note that the needed arguments are also changed.
- Renamed `check_tel_combination` to `get_stereo_events`, and changed the functionality to apply quality cuts and return a new data frame of stereo events with the `combo_type` and `multiplicity` parameters.
- Renamed `save_pandas_to_table` to `save_pandas_data_in_table`.
- Moved `get_stereo_events`, `get_dl2_mean` and `save_pandas_data_in_table` to `magicctapipe/io/io.py`.
- Moved `create_gh_cuts_hdu` to `magicctapipe/io/gadf.py`.

**lst1_magic_mc_dl0_to_dl1.py**
- Moved the custom containers to `magicctapipe/io/containers.py` and adapted the script to import them from the module.
- Pass only the pixels surviving the cleaning to the parametrization, which speeds up the process around 10%.
- **(EDIT)** Skip the process of the event whose cleaned image contains negative charges, since it fails the Hillas and timing parametrization.
- Skip the process of the event which failed to reconstruct finite timing parameters.

**magic_calib_to_dl1.py**
- The same updates as `lst1_magic_mc_dl0_to_dl1.py`.

**lst1_magic_event_coincidence.py**
- Renamed `load_lst_data_file` and `load_magic_data_file` to `load_lst_dl1_data_file` and `load_magic_dl1_data_files`, moved them to `magicctapipe/io/io.py`, and adapted the script to import them from the module.
- Changed the name of the setting from `type_lst_time` to `timestamp_type_lst`, which is the type of the LST timestamp. 
- Changed the name of the setting from `window_width (= 600 ns)` to `window_half_width (= 300 ns)`, which is the coincidence window half width.
- Changed the philosophy of the coincidence algorithm to scale all the timing-related parameters (`time_lst`, `time_magic`, `time_offsets`, `window_half_width`) in nanoseconds and use the `int64bit` format, instead of subtracting the UNIX time of an observation day and using float64bit type. I think this is a safer way in terms of a rounding issue, and now we don't need to round the parameters which happened everywhere in the script. Also I believe it improves the readability of the code.

**lst1_magic_stereo_reco.py**
- Use `loc` method instead of `query` to extract a shower event, since it speeds up the process around 20%.
- Skip the process of the event if it fails to reconstruct valid stereo parameters, which happens when the event contains images of width = 0 (though they should be removed by the default quality cuts).
- Don't save the stereo parameters which do not have finite values even if the reconstruction succeeds, i.e., `core_uncert`, `h_max_uncert`, `is_valid`. See [the definition](https://github.com/cta-observatory/ctapipe/blob/22655d189d6fe1330e01891ff12296b621313bdf/ctapipe/reco/hillas_reconstructor.py#L295-L307) of the reconstructor.

**lst1_magic_train_rfs.py**
- Moved `load_train_data_file` to `magicctapipe/io/io.py` and adapted the script to import it from the module.
- Removed `check_feature_importance` and added the functionality to the training functions.
- Removed `get_events_at_random` and added the functionality directly inside `train_event_classifier`.
- Renamed `train_direction_regressor` to `train_disp_regressor` as following the change of the regressor module from `DirectionRegressor` to `DispRegressor`.
- **(EDIT)** Renamed `load_train_data_file` to `load_train_data_files` and the argument from `input_file` to `input_dir` so that multiple MC files can be read.

**lst1_magic_dl1_stereo_to_dl2.py**
- Separated the functionality of `DirectionRegressor.predict` to `DispRegressor.predict` and `reconstruct_arrival_direction` functions to make it clear that `DispRegressor` only trains and predicts the DISP parameter, and the reconstruction of the arrival direction is done separately with the DISP method. Also, `DirectionRegressor` assumed that the input data is already separated per telescope combination type, but now `DispRegressor` does not assume it. So we can easily test training/applying RFs with the `combo_type` parameter without separating the RFs per telescope combination types.
- Don't transform telescope-wise Alt/Az direction to the RA/Dec coordinate, since for creating the IRFs/DL3 the RA/Dec direction transformed from the mean Alt/Az direction is used.

**lst1_magic_create_irf.py**
- Renamed `load_dl2_data_file` to `load_mc_dl2_data_file`, moved it to `magicctapipe/io/io.py`, and adapted the script to import it from the module.
- Removed `apply_dynamic_gammaness_cut` and `apply_dynamic_theta_cut`, and added the functionality to `create_irf`.

**lst1_magic_dl2_to_dl3.py**
- Removed `calculate_deadc` and added the functionality inside `load_dl2_data_file`.
- Moved `load_dl2_data_file`, `load_irf_files` to `magicctapipe/io/io.py`, and adapted the script to import them from the module
- Renamed `create_event_list`, `create_gti_table` and `create_pointing_table` to `create_event_hdu`, `create_gti_hdu` and `create_pointing_hdu`, changed the functionality to return the HDUs, moved them to `magicctapipe/io/gadf.py`, and adapted the script to import them from the module.
- Calculate the mean pointing Az direction in a range where the standard deviation of the pointing direction is smaller, 0 < az < 360 deg or -180 < az < 180 deg, which was originally suggested by @jsitarek in the pull request #81.


YoshikiOhtani:
I reprocessed the MCs produced by myself (Zd = 40 deg, Az = 90 deg) and the one night Crab data (2020-12-15, mid zenith) and reconstructed the spectrum, but it is somehow deviated from the reference curve so there must be some bugs in this branch, or already in the master branch by the previous pull request of the IRF interpolation. One buggy thing I already found is that there are differences between the original energy dispersion and the one from a DL3 file produced by the "nearest" method, which should be consistent. Thus, there must be an issue in `magicctapipe.io.load_irf_files`. I will check it as soon as possible, so @jsitarek and @aleberti, please wait a little bit more to check the code or reprocess the data with this branch, until I successfully reproduce a reasonable result. Sorry for the inconvenience!

YoshikiOhtani:
I think I found an issue in the pyirf function `pyirf.interpolation.interpolate_energy_dispersion`, not inside `magicctapipe.io.load_irf_files`. Since the master branch is already affected by this issue, I will create another issue and pull request to fix this.

magicctapipe/io/gadf.py
```diff
@@ -0,0 +1,306 @@
...
+        # Try to get the coordinate from the source name
+        source_coord = SkyCoord.from_name(source_name, frame="icrs")
+
+    except Exception:
+        # Use the input RA/Dec coordinate instead
+        logger.warning(
+            f"WARNING: The source name '{source_name}' could not be resolved. "
+            f"Setting the input RA/Dec coordinate ({source_ra}, {source_dec})..."
+        )
+        source_coord = SkyCoord(ra=source_ra, dec=source_dec, frame="icrs")
```
jsitarek, line 152: maybe add here an error message if from_name  failed and source_{ra,dec} is set to None?

YoshikiOhtani: Thanks for your comment. I pushed a commit to add this error message now.

YoshikiOhtani:
I finished reprocessing the data after fixing the interpolation issue, and now I got the consistent Crab spectrum by both MAGIC-only and LST-1 + MAGIC analyses. I will push one more commit to this branch to adapt the jupyter notebooks to the newer functions, but otherwise I'm satisfied with the changes. Please let me know if you have any other comments, thanks!

crab_magic
crab_lstmagic

magicctapipe/io/gadf.py
```diff
@@ -0,0 +1,312 @@
...
+            ("TIMEUNIT", "s"),
+            ("TIMESYS", "UTC"),
+            ("TIMEREF", "TOPOCENTER"),
+            ("ONTIME", on_time.value),
+            ("TELAPSE", elapsed_time.to_value(u.s)),
+            ("DEADC", deadc),
+            ("LIVETIME", effective_time.value),
+            ("OBJECT", source_name),
+            ("OBS_MODE", "WOBBLE"),
+            ("N_TELS", 3),
```
jsitarek, line 200: I guess this funciton can be also executed for MAGIC-only analysis, in which case this and next two lines would not be correct. 
I suppose that getting max of multiplicity could be included here. The chance that we have no 3-tel events in a run should be very small. 
More difficult to fill in the next two lines properly, because 2 telescopes can mean MAGIC-only, but also the recent M2+LST1 data

magicctapipe/io/io.py
```diff
@@ -0,0 +1,941 @@
...
+    for combo_type, (tel_combo, tel_ids) in enumerate(TEL_COMBINATIONS.items()):
+
+        multiplicity = len(tel_ids)
+
+        df_events = event_data_stereo.query(
+            f"(tel_id == {tel_ids}) & (multiplicity == {multiplicity})"
+        ).copy()
+
+        df_events["multiplicity"] = df_events.groupby(group_index).size()
+        df_events.query(f"multiplicity == {multiplicity}", inplace=True)
```
jsitarek, line 118: I do not understand this part. You already select the events by multiplicity in line 114, so why do you recompute multiplicity and apply the cut once again in lines 117-118?

magicctapipe/io/io.py
```diff
@@ -0,0 +1,941 @@
...
+        gammaness_weights = event_data["intensity"]
+
+    else:
+        raise ValueError(f"Unknown weight type '{weight_type}'.")
+
+    df_events = pd.DataFrame(
+        data={
+            "energy_weight": energy_weights,
+            "gammaness_weight": gammaness_weights,
+            "weighted_energy": np.log10(event_data["reco_energy"]) * energy_weights,
```
jsitarek, line 200: weighted_log_energy (??)

magicctapipe/io/io.py
```diff
@@ -0,0 +1,941 @@
...
+        If any DL1 data files could not be found in the input directory
+    """
+
+    # Find the input files
+    file_mask = f"{input_dir}/dl1_*.h5"
+
+    input_files = glob.glob(file_mask)
+    input_files.sort()
+
+    if len(input_files) == 0:
```
jsitarek, line 369: similar check can be also introduced for reading LST1 data in the function above

magicctapipe/reco/estimators.py
```diff
@@ -55,76 +53,81 @@ def __init__(
...
+        # Train a RF per telescope
+        tel_ids = np.unique(event_data["tel_id"])
 
         for tel_id in tel_ids:
 
-            df_events = event_data.query(f"tel_id == {tel_id}")
+            df_events = event_data.query(f"tel_id == {tel_id}").copy()
+            df_events.dropna(subset=self.features, inplace=True)
+
+            if self.use_unsigned_features:
```
jsitarek, line 74: you could write it slightly more compact 
```
x_train = df_events[self.features].to_numpy()
if self.use_unsigned_features:
    x_train = np.abs(x_train)
```

magicctapipe/reco/estimators.py
```diff
@@ -55,76 +53,81 @@ def __init__(
...
+            df_events = event_data.query(f"tel_id == {tel_id}").copy()
+            df_events.dropna(subset=self.features, inplace=True)
 
-        # Apply trained RFs per telescope:
-        tel_ids = np.unique(event_data.index.get_level_values("tel_id"))
+            if df_events.empty:
+                continue
 
-        for tel_id in tel_ids:
+            if self.use_unsigned_features:
```
jsitarek, line 119: as in the comment above this can be written in a slightly more compact way

magicctapipe/reco/estimators.py
```diff
@@ -195,212 +202,107 @@ def __init__(
...
+        # Train a RF per telescope
+        tel_ids = np.unique(event_data["tel_id"])
 
         for tel_id in tel_ids:
 
-            df_events = event_data.query(f"tel_id == {tel_id}")
+            df_events = event_data.query(f"tel_id == {tel_id}").copy()
+            df_events.dropna(subset=self.features, inplace=True)
+
+            if self.use_unsigned_features:
```
jsitarek, line 235: similar comment as above

magicctapipe/reco/estimators.py
```diff
@@ -195,212 +202,107 @@ def __init__(
...
-                )
+            df_events = event_data.query(f"tel_id == {tel_id}").copy()
+            df_events.dropna(subset=self.features, inplace=True)
 
-                dists_per_combo += np.array(theta.to(u.deg).value)
+            if df_events.empty:
+                continue
 
-            distances.append(dists_per_combo.tolist())
+            if self.use_unsigned_features:
```
jsitarek, line 279: also here

magicctapipe/reco/estimators.py
```diff
@@ -469,89 +374,98 @@ def __init__(
...
+        # Train a RF per telescope
+        tel_ids = np.unique(event_data["tel_id"])
 
         for tel_id in tel_ids:
 
-            df_events = event_data.query(f"tel_id == {tel_id}")
+            df_events = event_data.query(f"tel_id == {tel_id}").copy()
+            df_events.dropna(subset=self.features, inplace=True)
+
+            if self.use_unsigned_features:
```
jsitarek, line 395: one more

magicctapipe/reco/estimators.py
```diff
@@ -469,89 +374,98 @@ def __init__(
...
-        tel_ids = np.unique(event_data.index.get_level_values("tel_id"))
+            df_events = event_data.query(f"tel_id == {tel_id}").copy()
+            df_events.dropna(subset=self.features, inplace=True)
 
-        for tel_id in tel_ids:
+            if df_events.empty:
+                continue
 
-            df_events = event_data.query(f"tel_id == {tel_id}")
+            if self.use_unsigned_features:
```
jsitarek, line 439: one more

magicctapipe/scripts/lst1_magic/config.yaml
```diff
@@ -19,7 +19,7 @@ LST:
 
     increase_psf:
         use: true
-        smeared_light_fraction: 0.125
+        fraction: 0.125
```
jsitarek, line 22: shouldn't this be off now?

magicctapipe/scripts/lst1_magic/config.yaml
```diff
@@ -175,65 +175,60 @@ event_classifier:
...
+        min: null
+        max: null
 
 
 create_irf:
     quality_cuts: null
-    irf_type: 'software'   # select 'hardware', 'software', 'software_with_any2' or 'magic_stereo'
-    dl2_weight: null   # type of the weight for averaging tel-wise DL2 parameters, null, 'var' or 'intensity'
-    obs_time: 50   # used for weighting background events, unit: [hour]
+    irf_type: "software"  # select "software(_only_3tel)", "magic_only" or "hardware"
```
jsitarek, line 196: this (_only_3tel) in the comment, is a bit confusing, best to list each separately, software, software_only_3tel, ...

magicctapipe/scripts/lst1_magic/config.yaml
```diff
@@ -175,65 +175,60 @@ event_classifier:
...
-        cut_type: 'global'   # select 'global' or 'dynamic'
-        global_cut_value: 0.2   # used for the global cut
-        efficiency: 0.75   # used for the dynamic cut
-        min_cut: 0.1   # used for the dynamic cut
-        max_cut: 0.45   # used for the dynamic cut
+        cut_type: "dynamic"  # select "global" or "dynamic"
+        global_cut_value: "0.2 deg"  # used for the global cut
+        efficiency: 0.75  # used for the dynamic cuts
+        min_cut: "0.1 deg"  # used for the dynamic cuts
+        max_cut: "0.45 deg"  # used for the dynamic cuts
```
jsitarek, line 227: 0.45 deg is actually a lot, there is an overlap between ON and OFF region

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
```diff
@@ -2,17 +2,34 @@
...
+create only point-like IRFs, i.e., the effective area, energy migration
+and background model. If the input data is only gamma MC, it also skips
+the creation of the background model.
+
+There are 4 different IRF types which can be specified by the "irf_type"
+setting in the configuration file. The "hardware" type is supposed for
+the hardware trigger between LST-1 and MAGIC, allowing for the events of
+all the telescope combinations. The "software(_only_3tel)" types are
+supposed for the software coincidence with LST-mono and MAGIC-stereo,
+allowing for only the events triggering both M1 and M2. The "software"
```
jsitarek, line 15: maybe best to mention that MAGIC only events without LST1 are excluded for this "software" ones (unless we want to change it)

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
```diff
@@ -323,230 +84,371 @@ def create_irf(
...
+            unit=u.deg,
+        )
 
     extra_header = {
         "TELESCOP": "CTA-N",
         "INSTRUME": "LST-1_MAGIC",
         "FOVALIGN": "RADEC",
-        "PNT_ZD": (pnt_gamma[0], "deg"),
-        "PNT_AZ": (pnt_gamma[1], "deg"),
+        "PNT_ZD": (pnt_gamma[0].value, "deg"),
```
jsitarek, line 257: maybe safer: 
```
.to_value('deg')
```

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
```diff
@@ -323,230 +84,371 @@ def create_irf(
...
-        hdus_irf.append(hdu_rad_max)
+        irf_hdus.append(hdu_rad_max)
 
     else:
-        raise KeyError(f'Unknown type of the theta cut "{theta_cut_type}".')
+        raise ValueError(f"Unknown theta-cut type '{theta_cut_type}'.")
 
-    # Create an effective area HDU:
-    logger.info("\nCreating an effective area HDU...")
+    # Create an effective-are HDU
```
jsitarek, line 440: are ==> area

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
```diff
@@ -577,38 +479,39 @@ def create_irf(
...
+        irf_hdus.append(bkg_hdu)
 
-    # Save the data in an output file:
+    # Save the data in an output file
     Path(output_dir).mkdir(exist_ok=True, parents=True)
 
     output_file = (
-        f"{output_dir}/irf_zd_{pnt_gamma[0]}deg_az_{pnt_gamma[1]}deg"
-        f"_{irf_type}_{gam_cut_config}_{theta_cut_config}.fits.gz"
+        f"{output_dir}/irf_zd_{pnt_gamma[0].value}deg_az_{pnt_gamma[1].value}deg"
```
jsitarek, line 509: .to_value (?)

magicctapipe/scripts/lst1_magic/lst1_magic_dl1_stereo_to_dl2.py
```diff
@@ -2,82 +2,266 @@
...
+    # Then, we get the flip combination minimizing the sum of the
+    # angular distances between the head and tail candidates per shower
+    # event. In order to speed up the calculations, here we separate the
+    # input events by the telescope combination types.
+
+    reco_params = pd.DataFrame()
+
+    for tel_ids in TEL_COMBINATIONS.values():
+
+        multiplicity = 2 * len(tel_ids)
```
jsitarek, line 159: why do you multiply by 2 in here?

magicctapipe/scripts/lst1_magic/lst1_magic_dl1_stereo_to_dl2.py
```diff
@@ -2,82 +2,266 @@
...
+        # 4 combinations can be defined as follows:
+        #   [(head, head), (head, tail), (tail, head), (tail, tail)]
+        # where the i-th element of each tuple means the i-th telescope
+        # image. In case of 3 images we have in total 8 combinations.
+
+        flip_combinations = np.array(
+            list(itertools.product([0, 1], repeat=len(tel_ids)))
+        )
+
+        # Next, we define all the possible any 2 telescope combinations.
```
jsitarek, line 181: without "any"

magicctapipe/scripts/lst1_magic/lst1_magic_dl1_stereo_to_dl2.py
```diff
@@ -2,82 +2,266 @@
...
+
+            container = {}
+
+            # Set the directions of a given flip combination
+            for tel_id, flip in zip(tel_ids, flip_combo):
+                container[tel_id] = df_events.query(
+                    f"(tel_id == {tel_id}) & (flip == {flip})"
+                )
+
+            for tel_any2_combo in tel_any2_combinations:
```
jsitarek, line 202: wouldn't this work instead:
```
for tel_id_1,tel_id_2 in tel_any2_combinations:
```

magicctapipe/scripts/lst1_magic/lst1_magic_dl1_stereo_to_dl2.py
```diff
@@ -2,82 +2,266 @@
...
+                    lat2=u.Quantity(container[tel_id_2]["reco_alt"].to_numpy(), u.deg),
+                )
+
+                # Sum up the distance per flip combination
+                distances[i_flip] += theta.to_value(u.deg)
+
+        # Finally, we extract the indices of the flip combinations for
+        # each event with which the angular distances become minimum
+        distances = np.array(distances)
+        distances_min = distances.min(axis=0)
```
jsitarek, line 220: maybe you can use np.argmin

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
```diff
@@ -607,18 +147,27 @@ def dl2_to_dl3(input_file_dl2, input_dir_irf, output_dir, config):
...
     )
 
-    hdus.append(hdu_aeff)
+    hdus.append(aeff_hdu)
 
+    # Interpolate the energy dispersion with a custom way, since there
+    # is a bug in the function of pyirf v0.6.0 about the renormalization
     logger.info("Interpolating the energy dispersion...")
 
-    edisp_interp = interpolate_energy_dispersion(
```
jsitarek, line 614: any problem with the pyirf function that required rewriting it here?

magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
```diff
@@ -41,514 +55,357 @@
...
+                if np.count_nonzero(mask) == 1:
+                    n_coincidence[i_step] += 1
 
-                # Check the coincidence including only the right edge:
-                condition_btwn = np.logical_and(
-                    time_magic > time_lolim[i_evt],
-                    time_magic <= time_uplim[i_evt],
-                )
+                # Check the coincidence including only the right edge
+                mask_btwn = np.logical_and(cond_lolim_wo_eq, cond_uplim)
```
jsitarek, line 254: why do you use here a different condition? shouldn't the difference between < and <== be very small?

magicctapipe/scripts/lst1_magic/lst1_magic_stereo_reco.py
```diff
@@ -128,88 +112,91 @@ def stereo_reconstruction(
...
+        theta_uplim = u.Quantity(config_stereo["theta_uplim"])
 
-        if len(theta) == 0:
-            logger.info("--> No LST-1 + MAGIC events are found.")
+        if theta_max > theta_uplim:
+            logger.info(
+                f"--> The maximum angular distance is {theta_max.round(3)}, "
+                f"which is larger than the limit {theta_uplim}. Exiting..."
+            )
+            sys.exit()
```
jsitarek, line 179: maybe instead of exiting simply the events with this large offset can be filtered out. running over subruns it does not matter much, because we loose only 10s or so, but this check is problematic when running over whole runs. Removing the events should just clip the (sub)-run so I do not think there should be any problems with the effective time calculations later

magicctapipe/scripts/lst1_magic/lst1_magic_stereo_reco.py
```diff
@@ -218,123 +205,113 @@ def stereo_reconstruction(
...
-            # Set the stereo parameters to the data frame:
+            # Set the stereo parameters to the data frame
             params = {
-                "alt": stereo_params.alt.to(u.deg).value,
-                "alt_uncert": stereo_params.alt_uncert.to(u.deg).value,
-                "az": stereo_params.az.to(u.deg).value,
-                "az_uncert": stereo_params.az_uncert.to(u.deg).value,
-                "core_x": stereo_params.core_x.to(u.m).value,
-                "core_y": stereo_params.core_y.to(u.m).value,
-                "core_uncert": stereo_params.core_uncert.to(u.m).value,
```
jsitarek, line 280: was core uncertainty dropped on purpose?

magicctapipe/scripts/lst1_magic/lst1_magic_train_rfs.py
```diff
@@ -35,410 +39,306 @@
...
-    logger.info(f"Input file:\n{input_file}")
+    # Load the input files
+    logger.info(f"\nInput directory:\n{input_dir}")
 
-    data_train = load_train_data_file(
-        input_file,
-        config_rf["gamma_offaxis"]["min"],
-        config_rf["gamma_offaxis"]["max"],
-    )
+    offaxis_min = config_rf["gamma_offaxis"]["min"]
```
jsitarek, line 159: this part of getting offaxis min/max and converting to Quantity is used 3 times, maybe extract it into a function?

magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py
```diff
@@ -319,23 +297,26 @@ def magic_calib_to_dl1(
                 )
 
             else:
+                # The UNIX format with the "long" sub-format can get the
+                # timestamp without affected by a rounding issue
```
jsitarek, line 301: without being affected ... / without the rounding issue

magicctapipe/utils/functions.py
```diff
@@ -1,294 +1,317 @@
...
+    "calculate_off_coordinates",
     "transform_altaz_to_radec",
-    "check_tel_combination",
-    "save_pandas_to_table",
-    "get_dl2_mean",
-    "create_gh_cuts_hdu",
 ]
 
+# The geographic coordinate of ORM
+LON_ORM = -17.89064 * u.deg
```
jsitarek, line 26: those are also in io/gadf, so maybe better to import them from there

magicctapipe/utils/functions.py
```diff
@@ -1,294 +1,317 @@
...
+    ra_diff = pointing_ra - on_coord_ra
 
-    # Compute the wobble offset and rotation:
-    offsets = np.arccos(
-        np.cos(on_coord.dec)
-        * np.cos(tel_coord.dec)
-        * np.cos(tel_coord.ra - on_coord.ra)
-        + np.sin(on_coord.dec) * np.sin(tel_coord.dec)
+    # Calculate the wobble offset
+    wobble_offset = np.arccos(
```
jsitarek, line 256: why not simply use the angular distance function?

magicctapipe/utils/functions.py
```diff
@@ -1,294 +1,317 @@
...
+    wobble_rotation = Angle(wobble_rotation).wrap_at(360 * u.deg)
 
-    skyoffset_frame = SkyOffsetFrame(origin=on_coord, rotation=-mean_rot)
+    # Calculate the OFF coordinates
+    wobble_coord = SkyCoord(pointing_ra, pointing_dec, frame="icrs")
 
-    wobble_coord = SkyCoord(mean_offset, u.Quantity(0, u.deg), frame=skyoffset_frame)
-    wobble_coord = wobble_coord.transform_to("icrs")
+    rotations_off = np.arange(0, 359, 360 / (n_off_regions + 1)) * u.deg
+    rotations_off = rotations_off[rotations_off.to_value(u.deg) != 180]
```
jsitarek, line 273: why is 180 excluded? I would rather expect to exclude 0

magicctapipe/io/io.py
YoshikiOhtani: Indeed in line 114 we already extracted the events with a given multiplicity, but the problem is that with this condition the events not belonging to a given telescope combination are also extracted due to the condition `f"tel_id == {tel_ids}"`. For example, in case of `tel_ids == [1, 2]`, the condition in line 114 also extracts the events whose telescope multiplicity is 2 but belonging to other combination types (i.e., tel 1 events of the combination [1, 3], or tel 2 events of the combination [2, 3]), because the pandas query handles a list as "or" condition. Thus, it is needed to recalculate the multiplicity and apply the cut in line 117-118 to remove those events, whose multiplicity will be calculated to 1.

Actually I also thought that this part is confusing, but I did not come up with another clear way. So let me add some comments or explanations in the code. If you have some ideas please let me know.

magicctapipe/io/io.py
YoshikiOhtani: Thanks, I changed the parameter name with "log".

magicctapipe/reco/estimators.py
YoshikiOhtani: Thanks, I changed it as you suggested.

magicctapipe/reco/estimators.py
YoshikiOhtani: Thanks, changed.

magicctapipe/reco/estimators.py
YoshikiOhtani: Changed.

magicctapipe/reco/estimators.py
YoshikiOhtani: Changed.

magicctapipe/reco/estimators.py
YoshikiOhtani: Changed.

magicctapipe/reco/estimators.py
YoshikiOhtani: Changed.

magicctapipe/io/gadf.py
YoshikiOhtani: Thanks for your comment, indeed the keys are wrong in case of 2 tels observations. I changed the script to use the maximum multiplicity for `N_TELS` as you suggested. And for `TELLIST` and `INSTRUME`, I changed the script to assign proper instruments using the `combo_types` parameter. Now the instrument will be "LST1_M1_M2", "LST1_M1", "LST1_M2" or "M1_M2" depending on the data. It would be great if you could check the updated lines.

magicctapipe/scripts/lst1_magic/config.yaml
YoshikiOhtani: Yes, it should be OFF for processing all-sky MCs. Since it is now standard to use them, I disabled this method now. Also, I updated the default settings of `increase_nsb` method as same as those used by lstchain for processing Crab-path all-sky MCs.

magicctapipe/scripts/lst1_magic/config.yaml
YoshikiOhtani: I agree, actually I also thought the same thing. Now the `max_cut` is updated to 0.3 deg.

magicctapipe/scripts/lst1_magic/config.yaml
YoshikiOhtani: Thanks, I changed it.

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
YoshikiOhtani: Yes, this is already mentioned in the next line, but I slightly modified it.

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
YoshikiOhtani: I used `.value` because I'm sure that it is already in degree (defined in `load_mc_dl2_data_file`), but OK, I agree that `.to_value` would be safer, so I changed it.

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
YoshikiOhtani: thanks, I fixed it.

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
YoshikiOhtani: Changed.

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
YoshikiOhtani: Yes, please see the issue #94. I opened a pull request #95 to fix this issue separately, but it is already merged to this branch. If the pull request is merged to the master, this will not be the changes done by this pull request.

magicctapipe/scripts/lst1_magic/lst1_magic_dl1_stereo_to_dl2.py
YoshikiOhtani: It is because of the number of flips for each telescope event. The data frame has the multi index `[obs_id, event_id, tel_id, flip]`, so when we group per event the multiplicity becomes N_flip * N_tels.

magicctapipe/scripts/lst1_magic/lst1_magic_dl1_stereo_to_dl2.py
YoshikiOhtani: Removed.

magicctapipe/scripts/lst1_magic/lst1_magic_dl1_stereo_to_dl2.py
YoshikiOhtani: It does not work, because the `tel_any2_combinations` is a list containing tuples of any 2 tels combination, so we have to loop with one variable. In case of `tel_id = [1, 2, 3]`, the `tel_any2_combinations` becomes `[(1, 2), (1, 3), (2, 3)]`, so the `tel_any2_combo` will be `(1, 2)`, `(1, 3)` or `(2, 3)`.

magicctapipe/scripts/lst1_magic/lst1_magic_dl1_stereo_to_dl2.py
YoshikiOhtani: Thank you, I did not notice this function, which is useful actually, so I changed the code.

magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py
YoshikiOhtani: Thanks I changed it.

magicctapipe/utils/functions.py
YoshikiOhtani: Yes, I changed the script as you suggested.

magicctapipe/io/gadf.py
jsitarek: thanks, but where did you change it? this part still says 3 tels

magicctapipe/io/io.py
jsitarek: ok, I see now. A small explanation in the code would be useful

magicctapipe/io/gadf.py
YoshikiOhtani: Sorry, I have not yet pushed the commits, since I have not yet implemented your comments. But anyway let me push the commits now that I implemented so far.

YoshikiOhtani: Pushed them now.

magicctapipe/scripts/lst1_magic/lst1_magic_dl1_stereo_to_dl2.py
jsitarek: actually maybe it would be more clear instead of multiplying by 2 in line 159 divide by 2 in line 148 when the multiplicities are derived from group_size

magicctapipe/scripts/lst1_magic/lst1_magic_dl1_stereo_to_dl2.py
jsitarek: hmm, are you sure?
```
In [3]: tel_any2_combinations=[(1,2),(1,3),(2,3)]

In [4]: for tel_id_1,tel_id_2 in tel_any2_combinations:
   ...:     print(tel_id_1, tel_id_2)
1 2
1 3
2 3

In [5]: for (tel_id_1,tel_id_2) in tel_any2_combinations:
   ...:     print(tel_id_1, tel_id_2)
1 2
1 3
2 3
```

YoshikiOhtani: Oh really?? OK then sorry, it's my bad 😅. I will implement your suggestion.

magicctapipe/scripts/lst1_magic/lst1_magic_dl1_stereo_to_dl2.py
YoshikiOhtani: Thanks, I agree with you, so I implemented as you suggested.

magicctapipe/scripts/lst1_magic/lst1_magic_stereo_reco.py
YoshikiOhtani: Yes, it is because even if the stereo reconstruction is succeeded, the core uncertainty is not assigned to the container. Please see the definition of the [HillasReconstructor](https://github.com/cta-observatory/ctapipe/blob/22655d189d6fe1330e01891ff12296b621313bdf/ctapipe/reco/hillas_reconstructor.py#L295) in ctapipe (v0.12.0).

YoshikiOhtani: And also the `h_max_uncert` is also not assigned, thus I dropped the parameter.

magicctapipe/utils/functions.py
YoshikiOhtani: I guess you mean the `astropy.coordinates.angular_separation` function, so I changed the script to use it. I confirmed that the result is consistent with the one obtained by the equation.

magicctapipe/utils/functions.py
YoshikiOhtani: This is because, in this function the 180 deg is the same position as the ON source.

YoshikiOhtani: For more information this is how I understand. If we add 180 deg to the wobble rotation angle, the OFF coordinate is created at the same position as the source coordinate.
coordinates

YoshikiOhtani: I added a small comment on this code.

magicctapipe/scripts/lst1_magic/lst1_magic_stereo_reco.py
YoshikiOhtani: OK, I changed the script to do so. Thanks for your comment.

magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
YoshikiOhtani: The second condition (including only the right edge) is used for estimating the number of coincident events between the 100 ns offset steps. Since the final digit of the timestamps is 100 ns, the number of coincident events between the steps should be constant. And it can be estimated from each 100 ns step position by removing the left edge condition, because the offset scan is performed from smaller to larger values, and the events coincident with the left edge condition are the ones becoming not coincident between the steps. Those are useful to estimate the number of coincident events at the average offset, which is usually between the offset steps.

But I realized that the checks are not needed at the first scan, and if we remove those checks we could speed up the algorithm. I will update the code now.

YoshikiOhtani: Done. Now the speed becomes surprisingly faster by ~ 50%.

magicctapipe/scripts/lst1_magic/lst1_magic_train_rfs.py
```diff
@@ -35,410 +39,306 @@
...
+
+        # Adjust the number of training samples
+        if n_events_gamma > n_events_proton:
+            logger.info(f"Extracting {n_events_proton} events from the gamma MCs...")
 
-        elif n_events_bkg > n_events_gamma:
-            logger.info(
-                f"--> Number of background events is adjusted to {n_events_gamma}"
+            multi_indices_random = pd.MultiIndex.from_tuples(
+                tuples=random.sample(multi_indices_gamma.tolist(), n_events_proton),
```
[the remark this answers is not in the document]
YoshikiOhtani: OK, I set the seed 1000 in the beginning of the script.

magicctapipe/scripts/lst1_magic/lst1_magic_train_rfs.py
YoshikiOhtani: Thanks for the comment, I changed the code to do so.

magicctapipe/scripts/lst1_magic/config.yaml
YoshikiOhtani: Please let me know if you think that 0.3 deg is OK or not.

magicctapipe/utils/functions.py
YoshikiOhtani: For some reasons importing from io/gadf clashes the pipeline, so I changed the code to import them from `magicctapipe.utils.functions`.

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
YoshikiOhtani: This issue has just now been fixed in the master branch, so let me resolve it.

magicctapipe/scripts/lst1_magic/lst1_magic_stereo_reco.py
jsitarek: thx for the explanation, I was precisely wondering that it was the only one dropped.I checked also the current version of ctapipe, and while it added some new fields (like the tilted frame core and mean intensity) those uncertainties does not seem to be there  still, so best to skip them as you did

YoshikiOhtani: OK so just in case let me add a comment why we skip those parameters.

magicctapipe/utils/functions.py
jsitarek: Ok, now I see, I thought that you were defining wobble_rotation differently

jsitarek:
Thinking more about this I think it is best if the remaining refactoring
goes into this PR and the full enclosure IRF, as new development go into a
separate PR.

Cheers,
Julian

śr., 28 wrz 2022, 17:03 użytkownik Yoshiki Ohtani ***@***.***>
napisał:

YoshikiOhtani:
Hi @jsitarek, OK thanks, actually I was also thinking about that, so I will create another pull request later.

YoshikiOhtani:
OK so I have done. There are really minor changes so I don't think they need to be reviewed thoroughly. If there is no objection, let me merge it to the master branch.

aleberti:
Hi @YoshikiOhtani, yes, please go ahead and merge.

YoshikiOhtani:
OK, I will merge it now.

Issue #94, opened 2022-09-09
Interpolation of the energy dispersion matrix


YoshikiOhtani:
I found that the interpolated energy dispersion matrix with the "nearest" method is somehow changed from the original matrix as shown in the attached plots. It seems to me that there is an issue in the pyirf (v0.6.0) interpolation function `pyirf.interpolation.interpolate_energy_dispersion`, where the interpolated matrix is renormalized:
https://github.com/cta-observatory/pyirf/blob/3f9cfba672c8a25eb5094d25acf411c6b2104112/pyirf/interpolation.py#L92

According to the comment it tries to renormalize along the migration axis, but I think the specified axis is wrong. It seems the function assumes that the returned matrix `matrix_interp` from the `astropy.interpolate.griddata` is the shape (n_energy_bins, n_migration_bins, n_fov_offset_bins), but when I checked it by myself it is returned with an extra axis at the beginning which is originally used for storing multiple matrixes of all the grid points. So the current normalization with the axis 1 must be done along the energy axis, resulting the matrix different from the original one.

![edisp_before_interp](https://user-images.githubusercontent.com/39613098/189316602-12c250d7-d657-445d-887a-19cddeb642f8.png)
![edisp_after_interp](https://user-images.githubusercontent.com/39613098/189316594-07158b56-1154-4db8-bd2f-2e590807a419.png)

The wrongly interpolated energy dispersion clearly affects the reconstruction of the Crab spectrum ~~(I will add the plots later here)~~ as shown in the attached plots. Since the interpolation method is already updated in pyirf v0.7.0, it does not make sense to open an issue in the pyirf repository. One solution is to implement a custom interpolation method in our pipeline which normalizes along the migration axis, until we update pyirf to newer versions. Please let me know what you think @jsitarek.

![crab_sed](https://user-images.githubusercontent.com/39613098/189325009-08322466-96ad-4120-88c5-b6a9c77feecd.png)
![crab_sed_interp](https://user-images.githubusercontent.com/39613098/189325016-0ebb62e4-1dab-4e98-bf82-6db5e60da426.png)

jsitarek:
HI @YoshikiOhtani 

sorry for a late feedback.
Indeed you are right, the old code was wrong. The new code also changed the API, so I think your proposed solution is the best: we keep with the reimplemented function in the code, and refactor it once again to a new function when we update the pyirf  version. 
We can close this issue when #91 is merged

[...]

PR #95, opened 2022-09-09
Fix the issue of energy dispersion interpolation

YoshikiOhtani:
This pull request fixes the issue #94 about the interpolation of the energy dispersion. With this solution the renormalization is performed along the migration axis. I confirm that this fix works for my Crab samples.


YoshikiOhtani:
Let me merge this branch to the master since we agreed in the issue #94. It closes #94 .

PR #99, opened 2022-09-29
Create the full-enclosure IRFs

YoshikiOhtani:
This pull request implements a functionality to create the full-enclosure IRFs, i.e., the effective area, energy dispersion, PSF table and background model. It depends on the number of FoV offset bins about which IRF type will be created, in case of 1 bin the point-like IRFs (only the effective area and energy dispersion) and in the other cases the full enclosure IRFs. With this classification it is possible to create the point-like IRFs with diffuse gamma MCs, which helps us to perform the point-like 1D spectral analysis even if there are only diffuse test samples. When the input is point-like gamma MCs, one FoV offset bin will be automatically created (the settings in the configuration will be ignored) around the true FoV offset, and then only the point-like IRFs will be created.

Regarding the DL2 to DL3 process, I updated the code to interpolate the full-enclosure IRFs. For the PSF table, since there is a bug in pyirf v0.6.0 about the renormalization as same as the issue #94, I implemented a custom way to renormalize along the source offset bins. For the background model, I implemented the same engine `griddata`, but I'm not sure if it is OK or not so just a preliminary implementation.

Finally let me summarize the other noticeable changes:
- The setting name `irf_type` is changed to `event_type`, since the old name is confusing with the newly-introduced type "POINT-LIKE" and "FULL-ENCLOSURE". Accordingly, the name of the header card is also changed from `IRF_TYPE` to `EVT_TYPE`.
- Handle ring-wobble MCs as point-like data, which is classified by the difference of minimum and maximum viewcone angle.
- Fix an issue that a proper error message does not appear when input IRF files are not consistent in the `dl2_to_dl3` process.
- Add the `fov_offset_bins` and `source_offset_bins` settings in the configuration file.

Let me assign @aleberti and @jsitarek for reviewers. It would be much appreciated if you could give me some comments or suggestions.

### DIFF
--- a/magicctapipe/io/io.py
+++ b/magicctapipe/io/io.py
@@ -938,10 +938,7 @@ def load_irf_files(input_dir_irf):
             # Remove the empty card
             extra_header.pop(key)
 
-        elif n_values != n_input_files:
-            raise RuntimeError(f"The configurations do not match, at least '{key}'.")
-
-        elif len(unique_values) > 1:
+        elif (n_values != n_input_files) or len(unique_values) > 1:
             raise RuntimeError(f"The setting '{key}' does not match.")
 
         else:
--- a/magicctapipe/io/io.py
+++ b/magicctapipe/io/io.py
@@ -941,7 +941,7 @@ def load_irf_files(input_dir_irf):
         elif n_values != n_input_files:
             raise RuntimeError(f"The configurations do not match, at least '{key}'.")
 
-        elif unique_values > 1:
+        elif len(unique_values) > 1:
             raise RuntimeError(f"The setting '{key}' does not match.")
 
         else:
--- a/magicctapipe/io/io.py
+++ b/magicctapipe/io/io.py
@@ -517,7 +517,7 @@ def load_train_data_files(
     return data_train
 
 
-def load_mc_dl2_data_file(input_file, quality_cuts, irf_type, dl2_weight_type):
+def load_mc_dl2_data_file(input_file, quality_cuts, event_type, dl2_weight_type):
     """
     Loads a MC DL2 data file for creating the IRFs.
 
@@ -527,12 +527,12 @@ def load_mc_dl2_data_file(input_file, quality_cuts, irf_type, dl2_weight_type):
         Path to an input MC DL2 data file
     quality_cuts: str
         Quality cuts applied to the input events
-    irf_type: str
-        Type of the IRFs which will be created -
-        "software" assumes the software coincidence with any2 events,
-        "software_only_3tel" uses only the 3-tels events,
-        "magic_only" assumes MAGIC-only stereo observations, and
-        "hardware" assumes the LST-1 + MAGIC hardware trigger condition
+    event_type: str
+        Type of the events which will be used -
+        "software" uses software coincident events,
+        "software_only_3tel" uses only 3-tel combination events,
+        "magic_only" uses only MAGIC-stereo combination events, and
+        "hardware" uses all the any2 combination events
     dl2_weight_type: str
         Type of the weight for averaging telescope-wise DL2 parameters -
         "simple", "variance" or "intensity" are allowed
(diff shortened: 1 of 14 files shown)


magicctapipe/io/io.py
```diff
@@ -619,19 +619,30 @@ def load_mc_dl2_data_file(input_file, quality_cuts, irf_type, dl2_weight_type):
...
+    min_viewcone_radius = sim_config["min_viewcone_radius"][0] * u.deg
+    max_viewcone_radius = sim_config["max_viewcone_radius"][0] * u.deg
+
+    viewcone_diff = max_viewcone_radius - min_viewcone_radius
+
+    if viewcone_diff < u.Quantity(0.001, u.deg):
+        # Handle ring-wobble MCs as same as point-like MCs
+        viewcone = 0 * u.deg
+    else:
+        viewcone = max_viewcone_radius
```
aleberti, line 631: I see that `SimulationConfigContainer` has a `diffuse` attribute. Couldn't we use that to tell if it is ring wobble/point-like or diffuse and assign the `viewcone` value accordingly?

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
```diff
@@ -2,27 +2,30 @@
...
-all the telescope combinations. The "software(_only_3tel)" types are
-supposed for the software coincidence with LST-mono and MAGIC-stereo,
+This script processes MC DL2 events and creates the IRFs. It can create
+two different IRF types based on the number of FoV offset bins,
+"POINT-LIKE" in case of one bin and "FULL-ENCLOSURE" in the other cases.
+The effective area and energy dispersion HDUs are created in case of the
+"POINT_LIKE" IRFs, and in addition the PSF table and background HDUs in
+case of the "FULL-ENCLOSURE" IRFs. When the input gamma MC is point-like
+data, it creates one bin around the true FoV offset and creates only the
+"POINT-LIKE" IRFs.
```
aleberti, line 12: Maybe you can add also here the detail that ring-wobble MCs will be treated as point-like MCs.

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
```diff
@@ -97,50 +107,99 @@ def create_irf(
...
+            f"\nMean true FoV offset: {mean_true_fov_offset}"
+            f"\n--> FoV offset bins: {fov_offset_bins}"
+        )
+
+    # Here we decide the IRF type based on the number of FoV offset
+    # bins, "POINT-LIKE" in case of 1 bin and "FULL-ENCLOSURE" in the
+    # other cases. It allows for creating the "POINT-LIKE" IRFs with
+    # diffuse gamma MCs by selecting a given FoV offset region.
+
+    is_point_like = len(fov_offset_bins) == 2  # 1 bin -> n_edges = 2
```
aleberti, line 173: here the description and the following line do not match. maybe you can define:

```
is_point_like = (len(fov_offset_bins) - 1) == 1
```

magicctapipe/io/io.py
YoshikiOhtani: Thanks, I checked a ring-wobble MC data, and the `diffuse` attribute is flagged as 1. I also checked that my point MC is `diffuse = 0`, and the all-sky training diffuse MC is `diffuse = 1`, so it cannot be used for the separation.

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
```diff
@@ -118,10 +120,10 @@ def dl2_to_dl3(input_file_dl2, input_dir_irf, output_dir, config):
         pnt_az_mean = pnt_az_wrap_180deg.mean().wrap_at(360 * u.deg).value
 
     target_point = np.array([pnt_coszd_mean, pnt_az_mean])
-    logger.info(f"\nTarget point (cosZd, Az):\n{target_point.round(5).tolist()}")
+    logger.info(f"\nTarget point (cosZd, Az): {target_point.round(5).tolist()}")
 
     # Prepare for the IRF interpolations
-    interpolation_method = config_dl3["interpolation_method"]
+    interpolation_method = config_dl3.pop("interpolation_method")
```
aleberti, line 126: why using `pop`? in that way the key-pair is removed from `config_dl3`

aleberti:
Just as a curiosity, did you try to use the point-like IRFs with dynamic theta cuts with gammapy 0.20(.1)? I remember I tried before the summer (when we were discussing about the gammapy upgrade), but at the time it was not working (but I do not think I found the reason at the time).

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
YoshikiOhtani: Thanks, a bit differently but I changed it.

aleberti:
> Hi @aleberti, yes I used it and it worked fine with gammapy v0.20.1. I did not need to change anything of my script, and it succeeded by just following the instruction. The notebook `check_sed_from_dl3` is already updated to be used by gammapy v0.20.*. Do you remember which step you failed?

It was when reading the theta2 cuts information, if I remember correctly. But since it works fine now, I assume the problem is solved.

YoshikiOhtani:
Since it did not receive any other comments, let me merge it to the master branch. Just in case, this branch worked for my analysis samples so I hope there is no big issue.

PR #100, opened 2022-09-29
fixed misleading message

jsitarek:
when training RF instead of maximum offset value minimum was shown instead


jsitarek:
a small fixed of confusing message.
@YoshikiOhtani  I will merge it tomorrow

YoshikiOhtani:
Thanks @jsitarek, I did miss that. Let me merge it now to the master.

PR #102, opened 2022-10-01
fixed a problem with non agreeing table sizes causing crash of the DL…

jsitarek:
…2 generation script

in multiplicity calculation division by 2 was missing in one place

I also changed float division to integer division, since those values should be always even


YoshikiOhtani:
Thank you @jsitarek for catching the cause of the error, I'm generally fine with your pull request, but I pushed a commit to define a new `group_size` variable, since the line length was larger than the black limit of 88 characters. I will merge it now.

YoshikiOhtani (approved):

PR #105, opened 2022-10-12
Remove the allow mono trigger option

YoshikiOhtani:
This pull request removes the `--allow-mono` option from the script `magic_calib_to_dl1.py`. It was introduced not to process the mono-trigger data unexpectedly, since we do not know the trigger settings from file names, but I found another way to remove them so it is not needed now. @aleberti just in case, does this change affect the pull request about the Hillas image comparison?

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -51,6 +51,6 @@ dependencies:
   - corsikaio
   - pip:
       - lstchain~=0.9.6
-      - ctapipe_io_magic~=0.4.6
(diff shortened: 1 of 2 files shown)


aleberti:
No, the image comparison uses stereo data, so we are not ingesting mono data for that. So you can safely merge.

YoshikiOhtani:
OK thanks, then I will merge it now.

PR #109, opened 2022-10-15
Use matrix operation instead of looping over every LST event

YoshikiOhtani:
Speed up the coincidence search by using matrix operation instead of looping over every LST-1 events.

### DIFF
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
@@ -222,36 +222,32 @@ def event_coincidence(input_file_lst, input_dir_magic, output_dir, config):
         # maximizing offset. Finally, we again check the coincidence at
         # the average offset and then save the coincident events.
 
-        n_events_lst = len(timestamps_lst)
-
-        n_coincidence = np.zeros(len(time_offsets), dtype=int)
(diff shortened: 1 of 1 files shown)


jsitarek:
Hi Yoshiki,

Did you check if the speed up is due to more optimized algorithm or simply
from using multiple cores used in the matrix operations? It is important to
understand, because the data will be processed mostly one core per file.

Julian

sob., 15 paź 2022, 15:02 użytkownik Yoshiki Ohtani ***@***.***>
napisał:

YoshikiOhtani:
Hi @jsitarek, I think it is due to the former reason. I am testing it in my local machine, and do not see the increase of CPU. I can test it in the IT cluster if it is actually speeded up by specifying just one core.

YoshikiOhtani:
I tested it in the IT cluster, and it actually speeds up the process with specifying just one core. Please see the attached plot for the comparison of the process times before and after the updates, with my one-night CrabNebula samples. By comparing the median values of those distributions, now the process time is faster more than twice. I will make this pull request open now, so please let me know if you have any comments.

process_time

YoshikiOhtani:
By the way here I use the offset scan region from -10 ns to 0 ns, but if we optimize the region based on the data-taking period (around offset -3.1 us before June 2021 and offset -6.5 after), the process will take only less than 1 minute.

YoshikiOhtani:
I will merge it now to the master branch. Please let me know if you face any problems.

PR #110, opened 2022-10-17
Keep non-coincident MAGIC-stereo events

### LINKED ISSUE #108: How to keep non-coincident MAGIC-stereo events?
The current pipeline does not keep MAGIC-stereo events which are not coincident with any LST-1 events, since they are thought to be mostly hadronic origin. However, in order to properly compare the trigger rates, the fractions of telescope combination types between real data and MCs, etc, it would be good to keep them.

One issue that I came up with is how to label observation and event IDs for those events. In case of real data, the pipeline currently assigns LST observation and event IDs to coincident MAGIC events, and use the IDs to group telescope-wise events. However, it is not possible to assign any LST event IDs to non-coincident MAGIC-stereo events, and assigning NaN values to those event IDs will crash the pipeline, so we need a solution for this problem.

I think using MAGIC observation and event IDs instead would be a good solution, because we keep only LST-1 events coincident with MAGIC events so it's possible to assign the MAGIC IDs to the LST-1 events. But in this way, it would be needed to also change the scripts of later analysis stages. Currently in real data, the `obs_id` and `event_id` variables indicate LST observation and event IDs, and there are also `obs_id_magic` and `event_id_magic` variables to store the MAGIC IDs. I think it is better to keep those names and roles in the future, because now real data events are put together per LST observation or subrun IDs (e.g., `dl1_LST-1_MAGIC.Run${LST_RunID}.${LST_SubRunID}.h5`), so the IDs of their file names should be consistent with the `obs_id` and `event_id` variables. However, in the later stages when we group telescope-wise events, it currently uses `obs_id` and `event_id`, which will not work for non-coincident MAGIC events because they are NaN. Thus, we need to change the scripts to use `obs_id_magic` and `event_id_magic` instead to properly group the events of any telescope combination types for real data.

It would be great if you could give me any thoughts on this solution, or any other ideas or suggestions are also very welcome. If we agree on this solution, I can start to push some commits to implement this solution.

YoshikiOhtani:
With this pull request we keep the MAGIC-stereo events non-coincident with any LST-1 mono events. Since they do not have LST observation/event IDs, the MAGIC IDs are assigned for the `obs_id` and `event_id` variables. It closes #108.

**(EDIT)** It is based on another pull request #109 which should be merged to the master at first.

### DIFF
--- a/magicctapipe/io/io.py
+++ b/magicctapipe/io/io.py
@@ -284,9 +284,6 @@ def load_lst_dl1_data_file(input_file):
         input_file, key="dl1/event/telescope/parameters/LST_LSTCam"
     )
 
-    event_data.set_index(["obs_id", "event_id", "tel_id"], inplace=True)
-    event_data.sort_index(inplace=True)
-
     # Add the trigger time differences of consecutive events
     event_data = add_delta_t_key(event_data)
 
@@ -300,9 +297,7 @@ def load_lst_dl1_data_file(input_file):
 
     # Exclude the events with duplicated event IDs.
     # ToBeChecked: if it still happens in recently processed data
-    event_ids, counts = np.unique(
-        event_data.index.get_level_values("event_id"), return_counts=True
-    )
+    event_ids, counts = np.unique(event_data["event_id"], return_counts=True)
 
     if np.any(counts > 1):
         event_ids_dup = event_ids[counts > 1].tolist()
@@ -317,6 +312,8 @@ def load_lst_dl1_data_file(input_file):
     # Rename the columns
     event_data.rename(
         columns={
+            "obs_id": "obs_id_lst",
+            "event_id": "event_id_lst",
             "delta_t": "time_diff",
             "alt_tel": "pointing_alt",
             "az_tel": "pointing_az",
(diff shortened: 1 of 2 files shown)


YoshikiOhtani:
@jsitarek, I checked the fractions of different telescope combinations with my one-night CrabNebula samples. Please see the plots of the fractions along the zenith angles. Here I applied only the intensity cut (> 50 ph.e.), so I would expect that whole events are hadronic origin. The blue points indicate the mean pointing directions of observation runs, and the black dashed lines are the results of simple linear regression to briefly expect the fractions of a given zenith angle. I compared the expected fractions in the low energy range (< 30 deg) with what you wrote in the performance paper for proton MCs, there seem to be a few % differences in all the combinations. We may need to understand if the differences can be explained by missing helium MCs, differences of Data/MC azimuth angles, or something. Anyway it would be great if you could cross-check the results!

fraction_comb

YoshikiOhtani:
Let me also merge it to the master branch, since the implementation is already completed. I would like to report the discrepancy of the fractions in the next meeting, so it would be great if we could discuss it there.

PR #113, opened 2022-10-23
Minor refactoring

YoshikiOhtani:
With this pull request I refactored the combined analysis scripts to simply the code and remove unnecessary processes, as following the previous major refactoring pull request #91. I also updated the docstrings and comments in the code so that people can understand the process more easily. There are no changes on the API so this pull request should be transparent to users. Some bugs and issues were actually found during the refactoring, but they are already fixed by the other pull requests that I recently opened. 

One important thing is that I implemented a function `magicctapipe.io.format_object` to pretty-print the configuration, instead of looping over every key and value everywhere in the scripts, and this function only works with python v3.8, with which the `sort_dict = False` argument can be used. I already updated the required python version with the pull request #119, so users must re-create the magic-cta-pipe environment to use the updated scripts. Sorry for the convenience.


YoshikiOhtani:
I reprocessed my Crab samples with the refactored code to test if it does not make any issues, and finally got consistent results before and after the refactoring. So now I think it is safe enough to merge this branch to the master, so let me do it now.

PR #119, opened 2022-10-26
Update the python version and some other packages

YoshikiOhtani:
With this pull request I updated the python version to 3.8 to use a new feature, and accordingly other packages. Removing the requirement on `bokeh` is needed to update the python version, and regarding `sphinx` I just let it follow the requirement of the other packages. For the other packages I just updated the versions. @aleberti could you please have a look? Thanks in advance.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -4,11 +4,11 @@ channels:
   - default
   - conda-forge
 dependencies:
-  - python=3.7
+  - python=3.8
   - pip
   - astropy
   - black
-  - bokeh=1.0
+  - bokeh
   - nbsphinx
   - ctapipe=0.12
   - gammapy=0.19.0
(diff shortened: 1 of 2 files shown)


aleberti:
As we discussed via Slack, if the environment installs fine, I guess there are no problems in merging.

aleberti:
Maybe about the lstchain version upgrade: does this affect us? We use lstchain here and there, so maybe we need to be sure it works as before (e.g. in `create_dl3_index_files.py` and maybe in `io.py`, where `add_delta_t_key` is used; I guess that at lower level, i.e. cleaning and muon analysis, nothing has been really changed, I hope, but we can check)

YoshikiOhtani:
Thanks @aleberti, it is a good point. I looked into the changes on lstchain and it actually affects our pipeline, at least we need to modify `create_dl3_index_files`. But this is beyond the purpose of this pull request, so I will get the lstchain-related versions to the original ones.

YoshikiOhtani:
The environment is created successfully, so I will merge it to the master branch.

PR #120, opened 2022-10-26
Add event numbers in dl2

jsitarek:
this is to add original event/run numbers to the events  created from get_dl2_mean function. 
It is useful to later on find the same events in earlier stage of the data format


YoshikiOhtani:
Hi @jsitarek, I think that the run/event IDs are assigned to the data frame, isn't it? Please find that with the requested code the information is duplicated. They are used for multi indices, but if you need to access them, you can do `event_data_mean.reset_index(inplace=True)`, or `event_data_mean.index.get_level_values("obs_id")`.

スクリーンショット 2022-10-26 21 33 43

jsitarek:
you are right @YoshikiOhtani I thought that they were missing, because in my scripts I was concatenating information from multiple files and now I realized that while doing so, I was ignoring the original index. So the change is not needed. I will close the PR

PR #127, opened 2022-11-25
allow to further merge already merged files

jsitarek:
modified slightly the script for merging the files to allow to merge again files that were already merged. I.e. to merge run1_to_100, run_101_to_200, run_201_to_300 into run1_to_300. I checked (on MC dl1 files) that it does not break the regular  behavior) and on merged dl2 files that it works. 
One small issue is that if you merge e.g. 1_to_1000 and 2000_to_3000 it will stay say 1_to_3000 even if it does not have all the runs in between, which might be in some occasions misleading, however this was already happening before and it is also possible to have a run missing


YoshikiOhtani (approved):
Hi @jsitarek, thanks for the change, and it looks fine to me. I also realized this issue, but I do not find another solution and since it does not harm the analysis, let's keep it as it is now, and change it when it causes really an issue.

PR #129, opened 2022-12-05
Exclude lst1 dl1 merged

### LINKED ISSUE #128: Memory issues when running event coincidence on merged LST-1 DL1 files
Last week Axel reported that he had some jobs for the event coincidence failing with a memory issue (too much memory requested). We found out the problem happened because the event coincidence was run over LST-1 merged DL1 files. In such a case, the number of events is too large when performing the coincidence, and it causes the script to exit with an error similar to:

```python
Checking the event coincidence...
Traceback (most recent call last):
  File "/fefs/aswg/workspace/alessio.berti/miniconda3/envs/magic-lst1/bin/lst1_magic_event_coincidence", line 8, in <module>
    sys.exit(main())
  File "/fefs/aswg/workspace/alessio.berti/miniconda3/envs/magic-lst1/lib/python3.8/site-packages/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py", line 452, in main
    event_coincidence(
  File "/fefs/aswg/workspace/alessio.berti/miniconda3/envs/magic-lst1/lib/python3.8/site-packages/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py", line 226, in event_coincidence
    cond_lolim = timestamps_magic.value >= times_lolim[:, np.newaxis].value
numpy.core._exceptions.MemoryError: Unable to allocate 586. GiB for an array with shape (4724405, 133284) and data type bool
```

Therefore, it would be ideal to check the input file and exit if it is a merged DL1 file.

aleberti:
Closes #128.

### DIFF
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
@@ -454,6 +454,7 @@ def main():
             "Please run the script on LST-1 DL1 subrun files, otherwise you may "
             "run into memory issues. Exiting."
         )
+        sys.exit(1)
 
     with open(args.config_file, "rb") as f:
         config = yaml.safe_load(f)
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
@@ -47,6 +47,7 @@ import argparse
 import logging
 import sys
 import time
+import re
 from decimal import Decimal
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


YoshikiOhtani (commented):
Hi Alessio, thanks for the implementation, and it looks fine to me. One small suggestion is to put the code inside the `event_coincidence` function for a safe side, so that it also stops when one directly executes it.

YoshikiOhtani:
Yukiho started the MAGIC-LST analysis and encountered the same problem. I think we should merge it, as my comments are very minor one. Do you agree, @aleberti?

YoshikiOhtani (approved):

jsitarek:
this was never merged. The coincidence code evolved a bit from the time it was implemented, so  I am not sure if the original memory problem is still there, but I guess so, because there is still this line with a 2D table of MAGIC vs LST events. In the automatic pipeline we run coincidence over subruns 
If nobody complains until next week, I will merge it.

PR #143, opened 2023-05-16
Update coincident event search using large shower events

SeiyaNozaki:
Since the time offset depends on the observation dates, we needed to tune the offset scan region if we use the current method. This PR adds an option (`all_combination`) to select the best time offset among all possible combinations using large shower events. For the moment, 50 (MAGIC) x 50 (LST) events are selected by default to compute all possible time offsets and count coincident events with an assumed offset.

### DIFF
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
@@ -60,7 +60,6 @@ import yaml
 from astropy import units as u
 from ctapipe.instrument import SubarrayDescription
 from magicctapipe.io import (
-    format_object,
     get_stereo_events,
     load_lst_dl1_data_file,
     load_magic_dl1_data_files,
@@ -142,10 +141,10 @@ def event_coincidence(input_file_lst, input_dir_magic, output_dir, config):
         pre_offset_search = config_coinc["pre_offset_search"]
 
     if pre_offset_search:
-        logger.info(f"\nPre offset search will be performed.")
+        logger.info("\nPre offset search will be performed.")
         n_pre_offset_search_events = config_coinc["n_pre_offset_search_events"]
     else:
-        logger.info(f"\noffset scan range defined in the config file will be used.")
+        logger.info("\noffset scan range defined in the config file will be used.")
         offset_start = u.Quantity(config_coinc["time_offset"]["start"])
         offset_stop = u.Quantity(config_coinc["time_offset"]["stop"])
 
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
@@ -164,6 +164,8 @@ def event_coincidence(input_file_lst, input_dir_magic, output_dir, config):
         profiles = pd.DataFrame(
             data={"time_offset": time_offsets.to_value("us").round(1)}
         )
+    elif search_method == "all_combination":
+        profiles = pd.DataFrame(data={"time_offset": []})
 
     # Arrange the LST timestamps. They are stored in the UNIX format in
     # units of seconds with 17 digits, 10 digits for the integral part
@@ -286,10 +288,6 @@ def event_coincidence(input_file_lst, input_dir_magic, output_dir, config):
             )
             time_offsets = u.Quantity(time_offsets.round(), unit="ns", dtype=int)
 
-            profiles = pd.DataFrame(
-                data={"time_offset": time_offsets.to_value("us").round(1)}
-            )
-
             n_coincidences = np.array(
                 [
                     np.sum(
@@ -385,7 +383,15 @@ def event_coincidence(input_file_lst, input_dir_magic, output_dir, config):
 
         event_data = pd.concat([event_data, df_lst, df_magic])
         features = pd.concat([features, df_feature])
-        profiles = profiles.merge(df_profile)
+
+        if search_method == "scan":
+            profiles = profiles.merge(df_profile)
+
+        elif search_method == "all_combination":
+            df_profile = df_profile.query(f"n_coincidence_tel{coincidence_id} >  1")
+            df_profile = df_profile.drop_duplicates()
+            profiles = profiles.merge(df_profile, on="time_offset", how="outer")
+            profiles = profiles.sort_values("time_offset", ascending=False)
 
     if event_data.empty:
         logger.info("\nNo coincident events are found. Exiting...")
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
@@ -254,12 +254,6 @@ def event_coincidence(input_file_lst, input_dir_magic, output_dir, config):
 
                 n_coincidences.append(n_coincidence)
 
-            if not any(n_coincidences):
-                logger.info("\nNo coincident events are found. Skipping...")
-                continue
-
-            n_coincidences = np.array(n_coincidences)
-
         # Instead of the scan method, "all combination" method  will find
         # the best time offset among all possible combinations of time offset
         # using a small fraction of datasets. First, N events are extracted from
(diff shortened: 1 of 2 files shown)


SeiyaNozaki:
The updated procedure is 
* Pre offset search
1. Large-intensity events (100 LST x 100 MAGIC) are extracted
2. Compute all possible time offsets
3. Find the best offset giving the largest numbers of coincident events
4. Define the time offset scan region +/- full window width (600 ns) around the best offset
* Time offset scan using this offset and full datasets
Since offset scan region can be different between M1 and M2, `profile` data frame is like below:
Screenshot from 2023-05-22 00-07-25

Since the final coincident event search is performed using full datasets, the same coincident events are extracted w/ or w/o this method. The only weak point is that it does not work if the number of coincident events is too low (<10 events), but I think it is negligible.

magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
```diff
@@ -27,13 +27,17 @@
...
-would be needed to tune the offset scan region depending on the date
-when data were taken. The reason of the shift is under investigation.
+Please note that the time offset depends on the date of observations 
+as summarized below:
+* before June 12 2021: -3.1 us
+* June 13 2021 to Feb 28 2023: -6.5 us
+* March 10 2023 to March 30 2023: -76039.3 us
+* after April 13 2023: -25.1 us
+By default, pre offset search is performed using large shower events.
+The possible time offset is found among all possible combination of
```
aleberti, line 37: combination --> combinations

magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
```diff
@@ -182,6 +183,121 @@ def event_coincidence(input_file_lst, input_dir_magic, output_dir, config):
         df_magic["timestamp"] = timestamps_magic.to_value("s")
         df_magic.drop(["time_sec", "time_nanosec"], axis=1, inplace=True)
 
+        # Pre offset search is performed to define the offset scan region.
+        # First, N events are extracted from largerst intensity events for LST and
```
aleberti, line 187: largerst --> largest

magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
```diff
@@ -182,6 +183,121 @@ def event_coincidence(input_file_lst, input_dir_magic, output_dir, config):
...
+        # MAGIC. Then, it counts a number of coincident events within a defined
```
aleberti, line 188: counts **the** number

magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
```diff
@@ -182,6 +183,121 @@ def event_coincidence(input_file_lst, input_dir_magic, output_dir, config):
...
+        # window after shifting all possible combination (N x N) of time offsets.
```
aleberti, line 189: combination --> combinations

magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
```diff
@@ -182,6 +183,121 @@ def event_coincidence(input_file_lst, input_dir_magic, output_dir, config):
...
+
+            cond_lolim = timestamps_magic >= time_lolim
+            cond_uplim = timestamps_magic <= time_uplim
+
+            mask_lst_obs_window = np.logical_and(cond_lolim, cond_uplim)
+            n_events_magic = np.count_nonzero(mask_lst_obs_window)
+
+            if n_events_magic == 0:
+                logger.info(f"--> No {tel_name} events are found. Skipping...")
+                continue
```
aleberti, line 210: here we continue to the next tel_id, but I can imagine that if for M1 there are no events fulfilling the condition above, then also M2 should not fulfill, right? Probably if for one of the telescope the number of events in the LST window is 0, we should exit from the script with some exception and write an informative message

magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
```diff
@@ -182,6 +183,121 @@ def event_coincidence(input_file_lst, input_dir_magic, output_dir, config):
...
+                timestamps_lst
+                <= timestamps_magic[mask_lst_obs_window][-1] + window_half_width
+            )
+            mask_magic_obs_window = np.logical_and(cond_lolim, cond_uplim)
+
+            if np.count_nonzero(mask_magic_obs_window) == 0:
+                logger.info(
+                    f"\nNo LST events are found around {tel_name} events. Skipping..."
+                )
+                continue
```
aleberti, line 237: Same comment as above should hold here

magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
```diff
@@ -182,6 +183,121 @@ def event_coincidence(input_file_lst, input_dir_magic, output_dir, config):
...
+            # time cut around MAGIC observations is applied to the LST data set.
+            cond_lolim = (
+                timestamps_lst
+                >= timestamps_magic[mask_lst_obs_window][0] - window_half_width
+            )
+            cond_uplim = (
+                timestamps_lst
+                <= timestamps_magic[mask_lst_obs_window][-1] + window_half_width
+            )
+            mask_magic_obs_window = np.logical_and(cond_lolim, cond_uplim)
```
aleberti, line 231: just to make the code more "homogeneous", since you already reuse cond_lolim and cond_uplim, you may reassign above time_lolim and time_uplim to timestamps_magic[mask_lst_obs_window][0] - window_half_width and timestamps_magic[mask_lst_obs_window][-1] + window_half_width respectively

magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
aleberti: On 2nd thought, since the coincidence script is run over MAGIC DL1 files, it means the loop will be on 1 telescope only, either M1 or M2, so we definitely need to exit if no events are found

aleberti (changes requested):
Thanks Seiya for the implementation! I left some comments for your attention. One thing: did you try to see as x-check if you get similar/same results if you use the "all combination" approach and the old one with a time offset equal to the one found in the new method? Looking at the code it should, but since it should be a quick check, if you can do it, it would be great!

SeiyaNozaki:
Thank you Alessio for the comments!

> One thing: did you try to see as x-check if you get similar/same results if you use the "all combination" approach and the old one with a time offset equal to the one found in the new method?

Yes! Here you can also see the difference in the results between the current method and the updated one ('all combination' approach) using 2962 LST subrun data taken in March 2023. 
image

magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
SeiyaNozaki: Here I simply copy&pasted the existing lines, so maybe @YoshikiOhtani can comment on that. I saw sometimes this warning message (LST run 12273 taken on March 18 2023) and only LST-M2 events are found. I will check later...

https://github.com/cta-observatory/magic-cta-pipe/blob/b9f82d982143008bd053690dc32c9de89db37a32/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py#L197-L199

SeiyaNozaki:
Following the comment by @YoshikiOhtani in the meeting, I changed the scan range to +/-3 x half_window_width (1.5 x full window width) around the offset the pre-search found. Using this offset scan range, I got completely the same results (time offset, n_coincidence_events) as the current method for all of LST subruns.

magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
YoshikiOhtani: Yes, I also saw that the coincidence happens in only one of the combinations, even if I input the M1 and M2 datasets to the script. Maybe this is due to the case that `calib_to_dl1` process fails for some subrun files of either M1 and M2, but the counterparts are still used in the coincidence search. Anyway, this is out of the scope of this pull request so I think we should open another issue page.

YoshikiOhtani: Regarding the second thought, we can input the M1 and M2 datasets to the script simultaneously with `--input-dir-magic`, and in that case the loop will be on the two telescopes.

YoshikiOhtani (approved):

aleberti (approved):

PR #144, opened 2023-07-19
Zenith interpol

jsitarek:
changes for source path interpolation:

- added a possibility to use node up to a given distance from the data (this is useful to select only the nodes on the raising or falling part of the source path)
- allowed two interpolation schemes cosZdAz - like it was before, and cosZd (in which the interpolation is done only over cosZd, but normally you would select with the option above the nodes with similar azimuth)

### DIFF
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
@@ -118,14 +118,15 @@ def dl2_to_dl3(input_file_dl2, input_dir_irf, output_dir, config):
     else:
         pnt_az_mean = pnt_az_wrap_180deg.mean().wrap_at("360 deg").to_value("rad")
 
-    distances = np.degrees(
+    distances = (
         angular_separation(
             pnt_az_mean,
             np.pi / 2 - np.arccos(pnt_coszd_mean),
             irf_data["grid_points"][:, 1],
             np.pi / 2 - np.arccos(irf_data["grid_points"][:, 0]),
         )
-    )
+        * u.rad
+    ).to("deg")
 
     scheme = config_dl3.pop("interpolation_scheme")
     if scheme == "cosZdAz":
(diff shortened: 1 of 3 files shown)


jsitarek:
@YoshikiOhtani @aleberti if you could have a look it would be great

@Elisa-Visentin - this change can have a bit of a clash with your tests PR, because it asks for more parameters at the IRF generation stage. To reproduce the old behaviour you need to add those two lines to the config file in the "dl2_to_dl3" part:
```
    interpolation_scheme: "cosZdAz" 
    max_distance: null
```

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
```diff
@@ -117,8 +118,39 @@ def dl2_to_dl3(input_file_dl2, input_dir_irf, output_dir, config):
...
-    target_point = np.array([pnt_coszd_mean, pnt_az_mean])
-    logger.info(f"\nTarget point in (cos(Zd), Az): {target_point.round(5).tolist()}")
+    distances = np.degrees(
+        angular_separation(
+            pnt_az_mean,
+            np.pi / 2 - np.arccos(pnt_coszd_mean),
+            irf_data["grid_points"][:, 1],
+            np.pi / 2 - np.arccos(irf_data["grid_points"][:, 0]),
+        )
+    )
```
aleberti, line 128: Here I would use `Angle` e.g. 

```python
distances = Angle(
        angular_separation(
            pnt_az_mean,
            np.pi / 2 - np.arccos(pnt_coszd_mean),
            irf_data["grid_points"][:, 1],
            np.pi / 2 - np.arccos(irf_data["grid_points"][:, 0]),
        ), unit=u.rad
    ).to(u.degree)
```

and then later

```python
max_distance = config_dl3.pop("max_distance") * u.degree
```

so that it is clear that `max_distance` is in degrees (and `distances` as well)

aleberti (approved):
Looks good to me. I just left one comment that in any case is not critical, and it is a matter of taste probably.

magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
jsitarek: I modified the code to use Quantity for angles

aleberti (approved):

PR #145, opened 2023-08-01
changed the gap size from 0.1s to 1 s

jsitarek:
this is used in the computation of the dead time, but the output is also stored in DL3 index file since load_dl2_data is normally run with a quality cut in disp_diff its rate of events is much smaller, and cutting everything >0.1 s removes a lot of real ontime (~40%). This does not have affect the spectra (probably), because gammapy is recomputing the ontime, but still produces a very confusing information inside the DL3 index files 1s should be small enough to remove real gaps.

### DIFF
--- a/magicctapipe/io/io.py
+++ b/magicctapipe/io/io.py
@@ -60,7 +60,7 @@ EFFECTIVE_FOCLEN_LST = 29.30565 * u.m
 
 # The upper limit of the trigger time differences of consecutive events,
 # used when calculating the ON time and dead time correction factor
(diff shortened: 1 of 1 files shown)


YoshikiOhtani (approved):
Hi @jsitarek, sorry for a late response. I think this problem was discussed in the pull request #52 but somehow it was merged with the limit 0.1 sec. Actually I also saw these strange time information in DL3 files and wondered why it happened, but I just overlooked it since we could get a reasonable Crab spectrum with them. Anyway, I agreed with changing the limit to 1 sec and so approve this pull request. Thanks!

PR #146, opened 2023-08-17
Concentration

jsitarek:
Concentration was computed for LST images (in lstchain), but not for MAGIC ones. When data from different telescopes were merged even the LST one was removed.
It was also not computed for MCs. 
Now the parameters are computed and added.
I had to use a trick with prefix to be able to save the parameters with the same name as LST uses. 

I need to test that it works also at the following stages of the analysis and that it does not cause any side effects, so it is still a draft PR


YoshikiOhtani:
Hi @jsitarek, I agree with the change, though it deviates from the rule that I followed. Actually, I try to use the naming convention that ctapipe uses, and that's why this pipeline reverts the LST-custom naming to default ones when reading LST-1 DL1 files (in the [load_lst_dl1_data_file](https://github.com/cta-observatory/magic-cta-pipe/blob/1f27c608eba88bbe8fc9eb3c7ee588a0c9b3b50c/magicctapipe/io/io.py#L330C4-L348C4) function).

However, I see that the default names for the concentration parameters are a bit ambiguous (cog, core, pixel) and confused with the stereo parameters (core_x, core_y), so it would be better to add the prefix "concentration_" to the parameters, as you did in this change and lstchain does.

aleberti:
@jsitarek is there anything missing for this PR? Probably some tests that you wanted to perform?

jsitarek:
sorry for a long cold PR. I checked if it still works with ctapipe 0.19 compatibility and if the information is propoagate at least up to stereo files for both data and MCs. Plus I added a small sanity check to unit tests. I think it is ready now for review and hopefully merge

aleberti (approved):
Looks fine!

PR #150, opened 2023-09-19
Update the history of the time offset between LST-1 and MAGIC

SeiyaNozaki:
After WRS re-synchronization, the time offset between LST-1 and MAGIC was changed to -6.2 us.
This PR just updated the history comment in the script and config file.

### DIFF
--- a/magicctapipe/scripts/lst1_magic/config.yaml
+++ b/magicctapipe/scripts/lst1_magic/config.yaml
@@ -67,6 +67,12 @@ MAGIC:
 
 
 event_coincidence:
+    # History of the time offset between LST-1 and MAGIC
+    # before June 12 2021: -3.1 us
+    # June 13 2021 to Feb 28 2023: -6.5 us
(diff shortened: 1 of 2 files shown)


aleberti (approved):

jsitarek:
thank you @SeiyaNozaki for the update and @aleberti for approving and merging it.
However, if we need to update this part again, since those are just comments, and the values are needed only for a sanity check in the new coincidence scheme, maybe it is enough to include them only in one place instead of two (config file and the coincidence script)

SeiyaNozaki:
Now I'm thinking of adding another file somewhere describing pre-defined values depending on the run number or timestamp. Then, if pre offset search fails to find the possible candidate (or `pre_offset_search` is deactivated), it will use the pre-defined values. Do you agree? Then, we don't need to add corresponding lines to config file or header part of the coincidence script every power cycle. Just we need to add this info to this file and analyzer doesn't need to care about the values too much.

PR #151, opened 2023-09-25
Add license (BSD 3-Clause license).

aleberti:


aleberti:
> Not sure the static code check fails... offline it does not give any error.

solved by fixing numpy version.

setup.py
```diff
@@ -39,6 +39,7 @@
         'ctaplot~=0.5.5',
         'gammapy~=0.19.0',
         'uproot~=4.1',
+        'numpy<1.22.0a0',
```
jsitarek, line 42: I think it would be better to have the same condition in environment and setup.py, because now they ask for 1.21 or for <1.22.0a0 (both can be fulfilled simultaneously obviously, but why to limit it more in one place than in the other?

jsitarek (approved):
the PR has also some fixes on CI, and I have a small comment to it, but it can also stay like this if there is a reason for it.

setup.py
aleberti: and yes, I had to make some fixes on the CI because it gave some errors that before it did not give (e.g. when merging #141 ).

aleberti:
ok, now 1 test out of 60 is failing, and I would say I am "happy", in the sense that the test that was failing for me offline, now it fails here too (no idea why it was not failing before in the other branch, maybe the fixing of `numpy` version is the culprit?). I would say we can merge anyway, what do you think @jsitarek ? Then we can fix the file/test afterwards.

aleberti:
> ok, now 1 test out of 60 is failing, and I would say I am "happy", in the sense that the test that was failing for me offline, now it fails here too (no idea why it was not failing before in the other branch, maybe the fixing of `numpy` version is the culprit?). I would say we can merge anyway, what do you think @jsitarek ? Then we can fix the file/test afterwards.

correction: the test failing here is different, it is `test_load_train_data_files_off` while for me it was `test_load_train_data_files_g`.

jsitarek:
looking at the past few merged PRs everything was green with python 3.7 and started failing in mysterious places with 3.8. Now with a consistent error it is indeed better. Let's do like this: first merge the fix of the CI in the PR #152 (is it the same one as here??), and then we merge this thing about the licence from this PR. Afterwards we can work on fixing the remaining problem.

aleberti:
> looking at the past few merged PRs everything was green with python 3.7 and started failing in mysterious places with 3.8. Now with a consistent error it is indeed better. Let's do like this: first merge the fix of the CI in the PR #152 (is it the same one as here??), and then we merge this thing about the licence from this PR. Afterwards we can work on fixing the remaining problem.

well, not consistent error, the test failing is different from the one I had on my machine. Anyway yes, the CI changes are the same. If you approve the other MR, I then merge that and finally this one.

PR #152, opened 2023-09-25
Update the setup.cfg file.

aleberti:


aleberti:
Not sure the static code check fails... offline it does not give any error. Same as #151

aleberti:
Also here I had to update the CI and fix the `numpy` version. Still, one test out of 60 fails (same as in #151)

jsitarek (approved):

PR #153, opened 2023-09-25
Add pre commit

aleberti:
Add pre-commit configuration file and corresponding step in CI workflow.


aleberti:
One usual test failing, which sometimes fails, sometimes not. To be fixed, but in another PR.

PR #157, opened 2023-10-07
Tests

Elisa-Visentin:
More tests for the 4 LSTs pipeline

### DIFF
--- a/magicctapipe/conftest.py
+++ b/magicctapipe/conftest.py
@@ -189,6 +189,11 @@ def temp_coinc(tmp_path_factory):
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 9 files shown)


aleberti:
@Elisa-Visentin is this supposed to be merged after #170 ? or can it be merged before?

Elisa-Visentin:
Here I am implementing some tests while we are upgrading the pipeline. This will be merged with/after the next pipeline PR (i.e., upgrades of the RFs, Dl2, DL3...). PR #170 is about semi-automatic scripts, so not dependent on tests.

aleberti:
ok, I see. thanks!

PR #154, opened 2023-09-26
MCP upgrades: semi-automatic scripts and update of the current scripts (1 LST -> 4 LSTs); Torino team

ranieremenezes:
Checking the differences between the two branches

PR #159, opened 2023-10-20
Add deployment of docs in CI

aleberti:
See title.


aleberti:
Converting to draft, I cannot install correctly dependencies needed for building the docs in the CI. At the moment I am bound to wait for #154 to be merged, I cannot update environment and stuff to ctapipe v0.19 otherwise it will break that PR. Then #153 can be merged, followed by #157 . I guess then the RF->DL3 PR should be next (together with the PR with the semi-automatic scripts?), and finally #148 and #149, and then this one. So, for the moment, I wait...

aleberti:
So this also means that I will produce the documentation site (https://magic-cta-pipe.readthedocs.io/en/latest/index.html) manually until this is merged.

aleberti:
It seems I managed to fix environment and stuff for this to work for both docs and tests. This can be eventually merged even before the other PR mentioned in the previous message.

PR #162, opened 2023-10-24
Add release drafter action.

aleberti:
See title.


aleberti:
Not sure why one of the test is failing now, since I did not change anything on that side...

PR #163, opened 2023-10-25
MCP cleanup

Elisa-Visentin:
Here we can delete the scripts/functions/modules... that are no more used (maybe not even maintained). Feel 'free' to delete/undelete files.  I started this branch from the Torino one, which will be merged in a few days (actually, this PR won't be ready before this merging)
It will be merged before the ctapipe 0.19 upgrade (so that we won't have to upgrade the files that we don't want to maintain). BTW, every file kept in the repository master should have to be up-to-date and should be maintained in the near future.

Maybe we can also delete the download_data.sh and the modules/functions imported only by the deleted scripts, but we can discuss this.


aleberti:
> Here we can delete the scripts/functions/modules... that are no more used (maybe not even maintained). Feel 'free' to delete/undelete files. I started this branch from the Torino one, which will be merged in a few days (actually, this PR won't be ready before this merging) It will be merged before the ctapipe 0.19 upgrade (so that we won't have to upgrade the files that we don't want to maintain). BTW, every file kept in the repository master should have to be up-to-date and should be maintained in the near future.
> 
> Maybe we can also delete the download_data.sh and the modules/functions imported only by the deleted scripts, but we can discuss this.

I do not understand why you started from the Torino branch, the PRs are somehow independent, and I guess you do not want to delete the scripts you just introduced in the other PR :) It is easier to start PRs from the current version of the master (then if you want to use an already existing branch for the PR to have less local branches, fine, but also there I suggest to have new branches, it is less error prone). Even if there is a PR to be merged soon, you can merge the master into the PR branch later. For example, now this branch is not in line with the Torino one, the last merge commit is missing.

aleberti:
Well, we first delete the ones we know are not related. I anyway expect the other PR to be merged within this week, even if Gabriel does not reply (discussed this morning with Julian)

PR #178, opened 2023-11-29
Update time offset history

SeiyaNozaki:
After the power cut and WRS synchronization, the time offset between MAGIC and LST1 was changed: -1583.8 us on Nov 15, 118.8 us since Nov 16.


SeiyaNozaki:
The result after the WRS manual synchronization on Nov 30 is also included

aleberti (approved):
Thanks @SeiyaNozaki ! The docs and tests are failing, but they are fixed in other PRs/branches. I merge this now.

PR #179, opened 2023-12-01
added new tests

Elisa-Visentin:
Added new tests to check if output files have been produced in each of the analysis step (should solve the issue of no failed tests in case data are not adequately processed by the scripts). @aleberti could this be a solution or should I change/add other tests?


Elisa-Visentin:
#173 should solve the doc issue, if ready to be merged (Seems ready to me), so that we can then work on this PR

aleberti:
#173 is not a PR to fix docs, and actually it is not ready, it still does not do what I originally wanted. The problem you have is different. From the doc generation output:

```
/home/runner/work/magic-cta-pipe/magic-cta-pipe/magicctapipe/utils/functions.py:docstring of magicctapipe.utils.functions.transform_altaz_to_radec:20: WARNING: py:obj reference target not found: astropy.coordinates.angles.Longitude
/home/runner/work/magic-cta-pipe/magic-cta-pipe/magicctapipe/utils/functions.py:docstring of magicctapipe.utils.functions.transform_altaz_to_radec:35: WARNING: py:obj reference target not found: astropy.coordinates.angles.Latitude
```

so something probably was changed/corrected at astropy level. You just need to change  `astropy.coordinates.angles.Latitude` with `astropy.coordinates.Latitude` and `astropy.coordinates.angles.Longitude` with `astropy.coordinates.Longitude` in `utils/functions.py` and the documentation should build fine.

Elisa-Visentin:
I meant this:

```
packages/setuptools_scm/_integration/setuptools.py:30: RuntimeWarning: 
ERROR: setuptools==56.0.0 is used in combination with setuptools_scm>=8.x

Your build configuration is incomplete and previously worked by accident!
setuptools_scm requires setuptools>=61
```

BTW, in the meanwhile i will fix the astropy issue, thanks!

aleberti:
That error is still not solved unfortunately in #173 . It is not critical, because the docs building is fine even if that error is present. I will try to make it work in the lstchain branch, maybe it is a problem of python version.

Elisa-Visentin:
Anyway, I produced this PR to try to fix the issue we discovered about tests (with HaST and lstchain 0.10): some tests now should fail if no output file has been produced in a step (the older tests won't fail, but we cannot test, e.g., the DL2 script if we have no DL1 data)

Elisa-Visentin:
the docs issue seems just a version 'inconsistency' in the setup (could be due to one or more dependencies that requires a too old setuptools version). Maybe the newer version could be forced (I thought your PR did this) or we can update everything we can update (new ctapipe -> new pyirf....) and see if it is automatically solved

aleberti:
@Elisa-Visentin I think I found a nicer way to do this (looking at how ctapipe handle this very same issue). Since the files are created by running `subprocess.run()`, one can store the result in a variable, which holds, among others, the return code of the process. One can assert if that value is equal to 0 (=process successful) before returning the output file. I forced `magic_calib_to_dl1` to exit with a return code of 1 and the tests depending on those files where run and returned ERROR (different from FAILED, but was reported correctly). 

An example:
 ```python
@pytest.fixture(scope="session")
def M2_l1_monly(temp_DL1_M_monly, dl0_m2, config_monly):
    """
    Produce a DL1 file
    """

    process_run = subprocess.run(
        [
            "magic_calib_to_dl1",
            f"-i{str(dl0_m2[0])}",
            f"-o{str(temp_DL1_M_monly)}",
            f"-c{str(config_monly)}",
            f"-m{int(maxmonly)}",
            "--process-run",
        ]
    )

    assert process_run.returncode == 0
    return temp_DL1_M_monly
```

**Comment by aleberti** [comment]
This works well also when exceptions are raised, because python will then give the generic exit code 1.

**Comment by Elisa-Visentin** [comment]
@aleberti I implemented them as separate tests to try, in the near future, a little upgrade: skip all the tests that are non-sense because files have not been produced in a previous step (e.g., skip DL2 tests if no DL2 file exist). But maybe we can do this even with your version, but it would be really useful if the tests returns a failure instead of error  (I'm playing a bit with pytest: a global variable and an 'if Failed' should let us save a lot of time)

**Comment by Elisa-Visentin** [comment]
@aleberti I tried to change a bit the structure to group tests and, in case of non-processed/deleted/moved files (so that we don't have all the files we should have), skip the related tests. Now only implemented for stereo MC tests in MAGIC-only analysis (just to try it). What about this?

BTW, we must check again the combo-types in this branch before the merge. It seems that something has been lost in the merge (But I will check this)

**Comment by aleberti** [comment]
Thanks @Elisa-Visentin . To understand better, with the change you made, the skipped tests will be included in the summary pytest creates at the end right?   Then, will failures and errors be still shown? Looking at the documentation, when using -r, the default is fE (failures+errors). However now it is xXs. At this point, maybe we can just use -ra (see https://docs.pytest.org/en/6.2.x/usage.html#detailed-summary-report). By chance, do you have an example of how the change works if we have some files that were not created and therefore trigger the skipping of some test?

**Comment by Elisa-Visentin** [comment]
Hi Alessio, here I did not produce DL1 protons:
image
image
 The test checking if there are two DL1 proton files in the due folder fails (F status) while the ones dependent on it (if stereo exixt and the class that run tests on stereo) are skipped (s status)

**Comment by aleberti** [comment]
Ok, then from my side I think that this PR is good. Let's discuss on wednesday how to proceed, but I think this PR can be merged already, unless you need to add anything else.

**Comment by Elisa-Visentin** [comment]
I have to 'duplicate' the same structure (classes + dependencies) for the other tests (both in test_io and in test_io_monly), but I will do this in few days if you agree on this new structure

**Review by aleberti — APPROVED** [review]

**Comment by Elisa-Visentin** [comment]
Something strange happened again with combos in the master. I am trying to fix it now

File: magicctapipe/io/tests/test_io.py [hunk 1]
```diff
@@ -19,619 +21,721 @@
...
-        data_s, _, _ = load_mc_dl2_data_file(str(file), "width>0", "software", "simple")
-        assert np.all(data_s["combo_type"] < 3)
+        for file in stereo_mc:
+            event_data = pd.read_hdf(str(file), key="events/parameters")
+            event_data.set_index(["obs_id", "event_id", "tel_id"], inplace=True)
+            event_data.sort_index(inplace=True)
+            data = get_stereo_events(event_data, config_gen, "intensity>50")
+            assert np.all(data["intensity"] > 50)
+
+    def test_load_train_data_files_p(self, p_stereo):
```
jsitarek, line 268: the test for protons and gammas look exactly the same, if so it should be possible to have just one and use e.g. @pytest.mark.parametrize to run it over both p_stereo  and gamma_stereo

magicctapipe/io/tests/test_io.py
```diff
@@ -19,619 +21,721 @@
...
+        """
+        Check dictionary
```
jsitarek, line 270: maybe write explicitely that you check the dictorionary of the combo_type naming

magicctapipe/io/tests/test_io.py
```diff
@@ -19,619 +21,721 @@
...
+        assert "true_event_class" not in data.columns
+
+    def test_load_train_data_files_tel_g(self, gamma_stereo, config_gen):
+        """
+        Check dictionary
+        """
+
+        events = load_train_data_files_tel(str(gamma_stereo[0]), config_gen)
+        assert list(events.keys()) == [1, 2, 3]
+        data = events[1]
```
jsitarek, line 332: any reason why you use second telescope for protons in the function above, and first for gammas here? If not and both can use the same you can again make just one function and run it for two parameters as I explained a few comments above

magicctapipe/io/tests/test_io.py
```diff
@@ -19,619 +21,721 @@
...
+    def test_load_train_data_files_off(self, gamma_stereo):
+        """
+        Check off-axis cut
+        """
+        events = load_train_data_files(
+            str(gamma_stereo[0]), offaxis_min="0.2 deg", offaxis_max="0.5 deg"
+        )
+        data = events["LST1_M1"]
+        assert np.all(data["off_axis"] >= 0.2)
+        assert np.all(data["off_axis"] <= 0.5)
```
jsitarek, line 301: in the various "cut" tests I think it would be good to check also that the returned list is not empty.

magicctapipe/io/tests/test_io.py
```diff
@@ -19,619 +21,721 @@
...
+        Check on LST DL1
+        """
+        for file in dl1_lst:
+            events, _ = load_lst_dl1_data_file(str(file))
+            assert "event_type" in events.columns
+            assert "slope" in events.columns
+            assert "az_tel" not in events.columns
+            events = events.reset_index()
+            s = events.duplicated(subset=["obs_id_lst", "event_id_lst"])
+            s1 = ~s
```
jsitarek, line 559: I think ```assert np.all (s==False)``` would work as well instead of those two lines, but fine to keep it as it is if you prefer the current version

magicctapipe/io/tests/test_io.py
```diff
@@ -19,619 +21,721 @@
...
+        """
+        Check on MAGIC DL1
+        """
 
+        events, _ = load_magic_dl1_data_files(str(merge_magic), config_gen)
+        assert list(events.index.names) == ["obs_id_magic", "event_id_magic", "tel_id"]
+        assert "event_id" not in events.columns
+        events = events.reset_index()
+        s = events.duplicated(subset=["obs_id_magic", "event_id_magic", "tel_id"])
+        s1 = ~s
```
jsitarek, line 594: same as my previous comment

magicctapipe/io/tests/test_io_monly.py
```diff
@@ -17,456 +19,564 @@
...
-def test_format_object():
-    """
-    Simple check on a string
-    """
-    str_a = "a{b[[xz,cde}"
-    str_b = format_object(str_a)
-    assert str_b == "a b  xzcde "
-
+class TestGeneral:
+    def test_format_object(self):
```
jsitarek, line 23: the same function is tested in test_io.py, so I do not think you need to test it again here

magicctapipe/io/tests/test_io_monly.py
```diff
@@ -17,456 +19,564 @@
...
-def test_save_pandas_data_in_table(temp_pandas, pd_test):
-    """
-    Check on pandas dataframe (before = after saving it)
-    """
-    out = temp_pandas / "pandas.h5"
-    save_pandas_data_in_table(pd_test, str(out), "abc", "event")
-    df1 = pd.DataFrame(np.array([[1, 2], [3, 4], [5, 6]]), columns=["a", "b"])
-    df = pd.read_hdf(str(out), key="event")
-    assert df.equals(df1)
+    def test_save_pandas_data_in_table(self, temp_pandas, pd_test):
```
jsitarek, line 31: same here, also the same function is already tested in test_io.py

magicctapipe/io/tests/test_io_monly.py
```diff
@@ -17,456 +19,564 @@
...
+            config=config_gen,
+            offaxis_min="0.2 deg",
+            offaxis_max="0.5 deg",
         )
+        data = events[2]
+        assert np.all(data["off_axis"] >= 0.2)
+        assert np.all(data["off_axis"] <= 0.5)
 
-        assert np.all(data_m["combo_type"] == 3)
+    def test_load_train_data_files_tel_exc(self, temp_train_exc, config_gen):
```
jsitarek, line 186: I think this test is also the same as in test_io.py - if the exception is raised due to missing files for 3tel analysis it should also work for MAGIC-only

magicctapipe/io/tests/test_io_monly.py
```diff
@@ -17,456 +19,564 @@
...
+        dl2_mc = [p for p in gamma_dl2_monly.glob("*")] + [
+            p for p in p_dl2_monly.glob("*")
+        ]
+        for file in dl2_mc:
+
+            data_m, _, _ = load_mc_dl2_data_file(
+                str(file), "width>0", "magic_only", "simple"
+            )
+
+            assert np.all(data_m["combo_type"] == 0)
```
jsitarek: that one should be now 3, right?

aleberti: already fixed.

Elisa-Visentin:
Thank you Julian, I will check this. Just a minute because I am a bit lost in the combinations (there is still something wrong). BTW, thanks Alessio, but go to rest :)

Elisa-Visentin:
Now the combinations are fixed (right?) and we can work on duplicates...

magicctapipe/io/tests/test_io_monly.py
Elisa-Visentin: removed

magicctapipe/io/tests/test_io_monly.py
Elisa-Visentin: removed

magicctapipe/io/tests/test_io.py
```diff
@@ -19,619 +21,721 @@
...
+            )
+
+    def test_telescope_combinations(self, config_gen, config_gen_4lst):
+        """
+        Simple check on telescope combinations
+        """
+        M_LST, M_LST_comb = telescope_combinations(config_gen)
+        LSTs, LSTs_comb = telescope_combinations(config_gen_4lst)
+        assert M_LST == {1: "LST-1", 2: "MAGIC-I", 3: "MAGIC-II"}
+        assert M_LST_comb == {
```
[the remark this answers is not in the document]
Elisa-Visentin: fixed

magicctapipe/io/tests/test_io.py
```diff
@@ -19,619 +21,721 @@
...
-                    "MAGIC-I": 4,
-                    "MAGIC-III": 5,
-                }
-            }
-        )
+        with pytest.raises(
+            Exception,
+            match="Number of telescopes found in the configuration file is 5. It must be 6, i.e.: LST-1, LST-2, LST-3, LST-4, MAGIC-I, and MAGIC-II.",
+        ):
+            check_input_list(config_check)
```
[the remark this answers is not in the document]
Elisa-Visentin: Oh yes I remember why I did this. This test check what can happen if we make a typo in the config file so that we have two tels with the same name (eg. LST-2 and LST-2 instead of LST-1 and LST-2). If you hard-code this kind of combination into the test script, the python interpreter will complain because this is a Python error; if you put into the yaml file you can verify if an exception is raised (yes, it is raised)

magicctapipe/io/tests/test_io_monly.py
Elisa-Visentin: removed

magicctapipe/io/tests/test_io.py
Elisa-Visentin: done

Elisa-Visentin:
I'm working on all your comments and suggestions. Hopefully, tomorrow we will be ready for the merge and the new tagged version for the school

magicctapipe/io/tests/test_io.py
Elisa-Visentin: done

magicctapipe/io/tests/test_io.py
Elisa-Visentin: done

magicctapipe/io/tests/test_io.py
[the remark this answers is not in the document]
jsitarek: I think it must depend on the actual version of interpretter, because in ipython I get this:
```
In [1]: d1 = {"a": 1, "b": 2, "a": 3}

In [2]: d1
Out[2]: {'a': 3, 'b': 2}
```

A dictionary cannot have two same keys, so what is probably happening is that it is reading the same key twice and updating it. If you print the dictionary I'm pretty sure you will have only once LST-2, and the test triggers not because there are two entries with the same name, but because there is one missing.

magicctapipe/conftest.py
```diff
@@ -1030,7 +1031,7 @@ def real_dl3_monly(real_dl2_monly, IRF_monly, temp_DL3_monly, config_monly):
 
 
 @pytest.fixture(scope="session")
-def real_index(real_dl3):
+def real_index(real_dl3, temp_DL3):
```
jsitarek, line 1034: the usage of temp_DL3 in here looks confusing. 
temp_DL3 executes a fixture that creates the output directory, but you create the index file not in this temp_DL3 directory, but in real_dl3, yet you return the path of the temporary directory. 
so later on in test_exist_index(real_index) you do not check the directory of "real_dl3", but the one of of temp_DL3. Either there is something wrong with this test, or it needs more explanations what you are really testing

magicctapipe/conftest.py
```diff
@@ -1045,7 +1046,7 @@ def real_index(real_dl3):
 
 
 @pytest.fixture(scope="session")
-def real_index_monly(real_dl3_monly):
+def real_index_monly(real_dl3_monly, temp_DL3_monly):
```
jsitarek, line 1049: similar like in the comment above

magicctapipe/io/tests/test_io.py
jsitarek: sorry, this actually triggered linter because I put unnecessary space between the name of the function and the bracket, it should be:
```assert np.all(s==False)```

Elisa-Visentin: No worries, I will run pre-commit after a couple of major fixes and it will get fixed :)

magicctapipe/io/tests/test_io.py
Elisa-Visentin: @pytest.mark.parametrize is failing (I need to call the p_stereo fixture both to execute the scripts and to return a couple of directories, whose first  element must be provided to the MCP function, so a bit tricky), but I managed to remove the duplicate test anyway

magicctapipe/io/tests/test_io.py
Elisa-Visentin: Done

magicctapipe/conftest.py
Elisa-Visentin: fixed

magicctapipe/conftest.py
Elisa-Visentin: fixed

magicctapipe/io/tests/test_io.py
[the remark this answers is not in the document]
Elisa-Visentin: I removed this section of the test because if we hard-code the dictionary pyflakes checks will fail. BTW, this was not a very important test

aleberti:
@jsitarek can you quickly check the latest changes by Elisa? I think they are fine and reply to your comments. After that she can merge this and I can make the new release.

jsitarek:
@aleberti - done
let's merge it

PR #184, opened 2024-01-09
fix to create HDU and OBS index files with lstchain0.10

jsitarek:
after removing the old copied lstchain code for creating those index files and using the current (lstchain 0.10) code instead the script stoped working because of a small change of API in the function. Note that since the function is taken from lstchain, in the header file of the output index files 'lstchain' will be written as the creation program

@aleberti or  @Elisa-Visentin, please have a quick look and if you are fine with those changes we merge them into the "lstchain" branch

### DIFF
--- a/magicctapipe/scripts/lst1_magic/create_dl3_index_files.py
+++ b/magicctapipe/scripts/lst1_magic/create_dl3_index_files.py
@@ -17,6 +17,7 @@ import time
 from pathlib import Path
 
 from lstchain.high_level import create_hdu_index_hdu, create_obs_index_hdu
+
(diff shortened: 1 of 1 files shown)


aleberti:
Looks fine to me. I just run pre-commit since it was failing, a black line was removed probably by mistake. Tests were failing due to a network error, not because of the changes from this branch.

Elisa-Visentin:
OK for me. I just launched again the tests so that we can check if something is failing (should not, but before merging we must check it)

aleberti:
I merge anyway even if one test is failing, again because of some network issue, since the other two were succesfull.

PR #185, opened 2024-01-18
Update python version for readthedocs (now failing).

aleberti:
This should fix the building of the documentation in the readthedocs server, since the python version now required by lstchain and ctapipe is at least 3.9.


aleberti:
Just checked and the build in readthedocs now works fine.

Issue #191, opened 2024-01-21
Simplify/improve magic_calib_to_dl1


aleberti:
As reported by @jsitarek is #148 , to keep it in mind for later:

1.  for https://github.com/cta-observatory/magic-cta-pipe/blob/master/magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py#L175

> but what happens if somebody uses different tel_ids in config file for HaST data?
maybe we can at least introduce a check that for HaST data lst-1 is 1 etc. ?
> also the same setting should be used not only for HaST, but for all MAGIC+LST-1 observations

2. for https://github.com/cta-observatory/magic-cta-pipe/blob/master/magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py#L203

> so if you have joint data but with a software trigger it will also only use MAGIC positions ?? something I do not understand here because the new code works with the 2019/20 data that are pre-HaST

3. for https://github.com/cta-observatory/magic-cta-pipe/blob/master/magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py#L354
> this conversion here from the old numbering to the new numbering is really confusing. It needs at least some explanations.
> we can simplify the code in another PR (just a single loop over tels_with_trigger and converting them into assigned_tel_ids)

aleberti:
Point 3 addressed in #192

aleberti:
Point 1 and 2 addressed in #193

aleberti:
Everything was addressed, I close this issue.

PR #192, opened 2024-01-21
Simplify reassignment of tel_ids.

aleberti:
See issue #191 , point 3.

### DIFF
--- a/magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py
+++ b/magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py
@@ -343,45 +343,24 @@ def magic_calib_to_dl1(input_file, output_dir, config, max_events, process_run=F
                     n_islands=n_islands,
                 )
 
+            tel_ids_new_assignments = {
+                1: assigned_tel_ids["MAGIC-I"],
+                2: assigned_tel_ids["MAGIC-II"],
+                3: assigned_tel_ids["LST-1"],
+            }
+
             # Reset the telescope IDs
-            if tel_id == 1:
-                event_info.tel_id = assigned_tel_ids["MAGIC-I"]  # MAGIC-I
+            event_info.tel_id = tel_ids_new_assignments[event_info.tel_id]
 
-            elif tel_id == 2:
-                event_info.tel_id = assigned_tel_ids["MAGIC-II"]  # MAGIC-II
-
-            # M1+M2 trigger
-            if event.trigger.tels_with_trigger == [1, 2]:
-                tels_with_trigger_magic_lst = [
-                    assigned_tel_ids["MAGIC-I"],
-                    assigned_tel_ids["MAGIC-II"],
-                ]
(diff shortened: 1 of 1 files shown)


aleberti:
@jsitarek can you have a look to this as well? Here I simplified the way the new tel_ids are assigned for each event via a dictionary, following your comment in the PR for ctapipe 0.19.

magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py
```diff
@@ -343,45 +343,24 @@ def magic_calib_to_dl1(input_file, output_dir, config, max_events, process_run=F
...
-            elif event.trigger.tels_with_trigger == [1, 2, 3]:
-                tels_with_trigger_magic_lst = [
-                    assigned_tel_ids["LST-1"],
-                    assigned_tel_ids["MAGIC-I"],
-                    assigned_tel_ids["MAGIC-II"],
-                ]
+            event_info.tel_id = tel_ids_new_assignments[tel_id]
+
+            for idx, tel_idx in enumerate(event.trigger.tels_with_trigger):
+                event.trigger.tels_with_trigger[idx] = tel_ids_new_assignments[tel_idx]
```
jsitarek, line 356: it is normally dangerous to modify an array over which you iterate, but since we modify it at the same position that is being read in that iteration, this should be safe.

magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py
```diff
@@ -343,45 +343,24 @@ def magic_calib_to_dl1(input_file, output_dir, config, max_events, process_run=F
...
 
             # encode tels_with_trigger as an int value
             # that can be decoded later as a binary
             # tels_with_trigger = sum_{tel_id} 2**tel_id
             # where tel_id is only for those triggered
             tels_with_trigger_binary_int = np.array(
-                [2 ** (tel_id) for tel_id in tels_with_trigger_magic_lst]
+                [2 ** (tel_id) for tel_id in event.trigger.tels_with_trigger]
             ).sum()
```
jsitarek, line 364: if you want to simplify it a bit you can drop the explicit loop:
tels_with_trigger_binary_int=np.sum(2**np.array(event.trigger.tels_with_trigger))

jsitarek (approved):
you can merge it like this if you want, but I put a suggestion for simplification of the summing part.

magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py
aleberti: I now skip this step and do it directly later with a list comprehension.

magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py
aleberti: Done.

PR #193, opened 2024-01-21
Write MAGIC-only subarray

aleberti:
This addresses points 1 and 2 of #191.


aleberti:
@jsitarek can you have a quick look to this? I think this is what you wanted i.e.:
- as default, both MAGIC and LST are written in the subarray. LST=LST-1 only, since this script is for real data only so having more LSTs does not make sense (yet?).
- I added another option to write in the subarray only the MAGIC telescopes if the pipeline is used for MAGIC only analysis

jsitarek (approved):
looks good to me, thanks

PR #205, opened 2024-02-16
Sort files and assign train and test files.

aleberti:
I noticed that train and test files for gamma and protons are different between my machine and the ones of the CI. The `glob` function returns files not in particular order (e.g. alphabetical), but according to `ls -U`. This can change system by system. That is why sometimes they fail and sometimes they pass. Now I apply a sorting, and assign train and test as in my machine, where the tests pass.


aleberti:
Two of the matrices fail, but because of a connection error, the ones failing before now pass. Merging for the moment. For the network errors, I may need to either catch the error and retry or implement the download file cache function internally and set a larger timeout

PR #206, opened 2024-02-17
Remove averaging of positions in MCs.

### LINKED ISSUE #135: Saved telescope positions in MC are "wrong"
The saved telescope positions are relative to the unweighted average of the 3 telescopes. This will result in positions which are not consistent with the reconstructed or true core positions stored event-wise. This is very confusing!

https://github.com/cta-observatory/magic-cta-pipe/blob/dbcb3442683b26d7a03fb3be72e375eb199d5e72/magicctapipe/scripts/lst1_magic/lst1_magic_mc_dl0_to_dl1.py#L411-L418

aleberti:
Closes #135

### DIFF
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_mc_dl0_to_dl1.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_mc_dl0_to_dl1.py
@@ -29,7 +29,6 @@ from pathlib import Path
 
 import numpy as np
 import yaml
-from astropy import units as u
 from astropy.coordinates import Angle, angular_separation
 from ctapipe.calib import CameraCalibrator
 from ctapipe.image import (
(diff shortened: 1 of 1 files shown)


jsitarek (commented):
the code was also pruning the subarray of the telescopes that are not in use (i.e. have assigned id of 0). For the MAGIC+LST1 and MAGIC-only I think this is fine because those subarrays are already created as they shoud. 
But for the simulations with multiple LSTs, or for MAGIC-mono case (i.e. e.g. M2+LST-1 observations), the code I think will introduce changes w.r.t. previous behaviour. Might be useful if @ranieremenezes and @Elisa-Visentin have a look as well.

aleberti:
@jsitarek This is still done. The relevant part is:

```python
event_source = EventSource(
        input_file,
        allowed_tels=list(
            filter(lambda check_id: check_id > 0, assigned_tel_ids.values())
        ),  # Here we load the events for all telescopes with ID > 0.
        focal_length_choice=focal_length,
    )
```

by passing `allowed_tels`, internally the `SimTelEventSource` will select the telescopes from the subarray, and the lambda function is already taking away those with telescope ID set to. See https://github.com/cta-observatory/ctapipe/blob/main/src/ctapipe/io/simteleventsource.py#L696-L697

PR #209, opened 2024-02-22
Use the impact distance calculation from ctapipe.

aleberti:
Until now, we were using a function internal to MCP to calculate the impact distance. Since ctapipe provides it and it is calculated when calling the `HillasReconstructor`, it makes sense to use that.


aleberti:
As a cross-check, find attached the (real data) stereo files from the tests created with the current version from the master and the version from this branch. The calculated impact distances are consistent:

```python
In [1]: import pandas as pd

In [2]: import numpy as np

In [3]: df_orig = pd.read_hdf("dl1_stereo_MAGIC_LST-1.Run15337.0002_original.h5", key="events/parameters")

In [4]: df_new = pd.read_hdf("dl1_stereo_MAGIC_LST-1.Run15337.0002.h5", key="events/parameters")

In [5]: impact_orig = np.array(df_orig["impact"])

In [6]: impact_new = np.array(df_new["impact"])

In [7]: np.allclose(impact_new, impact_orig, equal_nan=True)
Out[7]: True
```

aleberti:
@jsitarek can you quickly review this PR? It is here since a long time

jsitarek:
sorry for missing this one, and thanks for merging it @aleberti

PR #210, opened 2024-02-28
Handle loading LST DL1 file from lstchain v0.9

gabemery:


Elisa-Visentin:
Fine for me. Here you are now assuming LST-1 ID = 1, so you loose a bit the IDs generalization, but we can accept this and fix it in the future by importing the configuration file (BTW, real data now only have ID =1)

aleberti:
Well, it can be fixed now I would say. We can just simply pass the dictionary of the mc tel ids (not all the config, unless we need something else from there to be passed to the function), so that even if there are more LSTs added in the future in the real data, the function is already usable.

Elisa-Visentin:
But I wanted to spare him some extra work both on the coincidence script and on the tests. He has already offered to fix the SubArray!

gabemery:
@aleberti This fix is meant to handle loading dl1 data from existing lstchain v0.9 files. So it will never be applied to dl1 data from LST2-4.

In case a dl1 file contains the data from multiple telescopes (not sure if this is the plan) and such a fix is still needed in the future, the tel_ids could probably be read from the dl1 file.

aleberti:
Well, in any case the ID is changed later in the coincidence script. Also, we cannot even get the telescope ID from the broken subarray. So, not much to do anyway.

aleberti:
There are no new comment since 2 weeks. @jsitarek green light from your side to merge? You have been assigned as 2nd reviewer.

magicctapipe/conftest.py
```diff
@@ -382,6 +384,22 @@ def dl1_lst(base_url, env_prefix):
     return LST_dl1
 
 
+@pytest.fixture(scope="session")
+def dl1_lst_old(base_url, env_prefix):
+    LST_dl1_old = []
+    for file in DL1_LST_old_lstchain:
```
jsitarek, line 390: since it is only one file, and I do not think we want to add more in the future, we could have skipped the loop and define the file as simple string rather then list

jsitarek (approved):
I suggested one possible simplification, but since this is in the tests, it is also perfectly fine with me if you merge it as it is. Sorry for a big delay in checking this PR

jsitarek:
thx @Elisa-Visentin 
as soon as the checks finishes correctly I think we can merge

PR #218, opened 2024-04-24
Add script for checking availability MAGIC _Y_ runs.

joannaWojS:


magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,88 @@
+#This script allows us to get information about every MAGIC run ID (and subruns) in files used for common data analysis (MAGIC1, MAGIC2, #LST1). You can also run the script using Jupyter Notebook.
```
jsitarek, line 1: why # before LST?
remove "You can also run the script using Jupyter Notebook."

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,88 @@
...
+
+#The MAGIC files that can be used for analysis are located here:
```
jsitarek, line 3: ... are located in the IT cluster in the following directory:

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,88 @@
...
+#/fefs/onsite/common/MAGIC/data/M{tel_id}/event/Calibrated/{year}/{month}/{day}
+
+#In this path, 'tel_id' refers to the telescope ID, which must be either 1 or 2. 'Year,' 'month,' and 'day' specify the date.
+
+#In the first step, we have to load a dataframe that contains information about the date, the name of the source, and the range of MAGIC #runs. The file in file_path was generated using the spreadsheet (Common MAGIC LST1 data) from the following link:
+
+#https://docs.google.com/spreadsheets/d/1Tya0tlK-3fuN6_vXOU5FwJruis5vBA9FALAPLFHOfBQ/edit#gid=1066216668
```
jsitarek, line 10: better to remove the link. It includes private MAGIC and LST information, while github is public. The link is protected from random access, but still better to avoid giving its link directly

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,88 @@
...
+
+import pandas as pd
+from datetime import datetime, timedelta
+import os
+import re
+
+file_path = '/fefs/aswg/workspace/joanna.wojtowicz/data/magic_first_and_last_runs.csv'
```
jsitarek, line 17: this is not directly the .h5 file of Federico, so if you want to use it, you should also add a scripts that generates it from Federico's file

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,88 @@
...
+df = pd.read_csv(file_path,sep='\t', dtype={'Date (LST conv.)': str, 'Source': str, 'First run': int, 'Last run': int})
+
+#df
```
jsitarek, line 20: remove

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,88 @@
...
+
+def check_run_ID(path, filename, first_run, last_run, date, source):
+    # Extract numbers from filename and check range
+    run_ids = [int(s) for s in re.findall(r'\d+', filename)]
```
jsitarek, line 24: this is a loop over all the numbers in the file name, and for MAGIC files this means also date, subrun, wobble angles, and often part of the name of the source. Or is it over LST runs?

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,88 @@
...
+    matched = False
+    magic_runs = []
+    
+    for id in run_ids:
+        if first_run <= id <= last_run:
+            matched = True
```
jsitarek, line 30: matched is not used

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,88 @@
...
+
+    path = f"/fefs/onsite/common/MAGIC/data/M{tel_id}/event/Calibrated/{year}/{month}/{day}"
+    
+
+    if os.path.exists(path):
+        files = os.listdir(path)
+        for filename in files:
+            if source in filename:
+                results = check_run_ID(path, filename, first_run, last_run, date, source)
+                #We will see many results becuse a file with a run ID has subruns.
```
jsitarek, line 61: because

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,88 @@
...
+    month = new_date[4:6]
+    day = new_date[6:8]
+    
+    results_count = {}
+
+    path = f"/fefs/onsite/common/MAGIC/data/M{tel_id}/event/Calibrated/{year}/{month}/{day}"
+    
+
+    if os.path.exists(path):
+        files = os.listdir(path)
```
jsitarek, line 57: You loop over all the files in the directory, so if there is some other file in this catalog (e.g. ssignal files) they will be considered as well as regular runs if they contain the run number in their name. You should use a regular expression to match the format of the _Y_ run:
date_M1/M2_run.subrun_Y_source[wobble string].root

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,88 @@
...
+                        results_count[result] += 1
+                    else:
+                        results_count[result] = 1
+    #else:
+        #print(f"No such file or directory: {date}")
+    
+    for result, count in results_count.items():
+        print(f"{result} \t {count}")
+    
+print(f'For the MAGIC 1 telescope:')
```
jsitarek, line 74: you can make a loop here over tel_id in [1,2]

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,88 @@
...
+                #We will see many results becuse a file with a run ID has subruns.
+                #We must count the same results to get information how many subruns we have.
+                for result in results:
+                    if result in results_count:
+                        results_count[result] += 1
+                    else:
+                        results_count[result] = 1
+    #else:
+        #print(f"No such file or directory: {date}")
+
```
jsitarek, line 70: we should add this listing of missing files

jsitarek (changes requested):
thanks for the script.
I put some comments in the code

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
joannaWojS: Done.

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,86 @@
...
+file_path = '/fefs/aswg/workspace/joanna.wojtowicz/data/magic_first_and_last_runs.csv'
+df = pd.read_csv(file_path,sep='\t', dtype={'Date (LST conv.)': str, 'Source': str, 'First run': int, 'Last run': int})
+
+def check_run_ID(path, filename, first_run, last_run, date, source):
+    Y = f'_Y_{source}' 
+    #'Y' because we have to be sure that the function counts right filename.
+
+    if Y in filename:
+        # Extract numbers from filename and check range
+        run_ids = [int(s) for s in re.findall(r'\d+', filename)]
```
jsitarek, line 24: I still do not like how the run_ids are extracted: since you get all the numbers in the file name: date, numerical parts of the source name, subrun, wobble angles, etc., it might also happen that you get runid-like number in one of those other numbers.
Is filename the base name, or with the full path?
If it is the base name you can extract the run id (and only this):
```int(filename.split("_")[2].split(".")[0])```

File: magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py (hunk 7 above)
Comment (jsitarek) [reply magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py]: still not used

File: magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py [hunk 13]
```diff
@@ -0,0 +1,86 @@
...
+        matched = False
+        magic_runs = []
+    
+        for id in run_ids:
+            if first_run <= id <= last_run:
+                matched = True
```
jsitarek, line 30: matched is not used

joannaWojS:
> hi @joannaWojS just a reminder about this branch - it still has some comments to be implemented before merging

Hi, thanks for the reminder. I'll do it tomorrow when I get back from a conference.

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
joannaWojS: I have just added a filename expression as date_M(tel_id)_run.subrun_Y_source, because on the common MAGIC+LST1 database are observations that don't have a [wooble string] part in a filename on the IT cluster. (example: 2021_05_13) Therefore I am not sure whether we should skip or include the files?

joannaWojS:
> One question: is this supposed to be also a standalone script? i.e. included in the console script? I think it should be better, but I am not sure about the original intention.

A few words of explanation about the script: Another script, list_from_h5.py, prints a list of LST and MAGIC runs for a date/source from the config file. The script uses two h5 files - one for MAGIC and another for LST data. The script check_MAGIC_runs displays data that are then used in the h5 file. Additionally, the scripts display missing files that are included in Federico's database. As Elisa mentioned, this should be added to the other scripts for databases.

joannaWojS:
> Hi Joanna, I was running a test and I realized that, in 2021_03_17, there is a typo in the name of one of the sources in the database, switching X and S. Now I added a few lines to the list script, so that now only sources with the same name both in LST and in M1 and in M2 are added to the dat file of the sources to be processed (just to prevent crashes in case of typos or unmatching names), whereas the txt files are still produced according to the name of the sources in the respective database (so that we can try to find out errors by looking at the file names)

You're right. However, in the file Common_MAGIC_LST1_data_MAGIC_RUNS_subruns (located in the same folder on the IT cluster), the source has the correct name. I think this file should be used in the scripts because it also contains additional information about the number of subruns.

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,134 @@
...
+#/fefs/onsite/common/MAGIC/data/M{tel_id}/event/Calibrated/{YYYY}/{MM}/{DD}
+
+#In this path, 'tel_id' refers to the telescope ID, which must be either 1 or 2. 'YYYY', 'MM', and 'DD' specify the date.
+
+#In the first step, we have to load a dataframe that contains information about the date, the name of the source, and the range of MAGIC #runs. The file in file_path was generated using the spreadsheet "Common MAGIC LST1 data".
+
+import pandas as pd
+from datetime import datetime, timedelta
+import os
+import re
```
jsitarek, line 13: this line is pointed in pylint as imported but not used

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,134 @@
+#This script allows us to get information about every MAGIC run ID (and subruns) in files (in a time interval) used for common data analysis (MAGIC1, MAGIC2, LST1). 
+
+#The MAGIC files that can be used for analysis are located in the IT cluster in the following directory:
+#/fefs/onsite/common/MAGIC/data/M{tel_id}/event/Calibrated/{YYYY}/{MM}/{DD}
+
+#In this path, 'tel_id' refers to the telescope ID, which must be either 1 or 2. 'YYYY', 'MM', and 'DD' specify the date.
+
+#In the first step, we have to load a dataframe that contains information about the date, the name of the source, and the range of MAGIC #runs. The file in file_path was generated using the spreadsheet "Common MAGIC LST1 data".
```
jsitarek, line 8: new line missing (...#runs ...)

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,134 @@
...
+def existing_files( tel_id, date, source, magic_runs ):
+
+    magic_runs = str(magic_runs)  
+    date_obj = datetime.strptime(date, '%Y%m%d')
+    date_obj += timedelta(days=1)
+    new_date = datetime.strftime(date_obj, '%Y%m%d')
+    YYYY = new_date[:4]
+    MM = new_date[4:6]
+    DD = new_date[6:8]
+    Y = f"_Y_"
```
jsitarek, line 49: you need 'f' before the string only if you have some placeholder {...}, here it is not needed

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,134 @@
...
+    path = f"/fefs/onsite/common/MAGIC/data/M{tel_id}/event/Calibrated/{YYYY}/{MM}/{DD}" 
+    
+    if os.path.exists(path):
+        files = os.listdir(path)
+        count_with_source = 0  
+        count_with_run_id = 0
+            # Counter for files that include the source.  
+            # Counter for files that include the run_id.
+        for filename in files:
+            if date and source and Y in filename:
```
jsitarek, line 60: I do not think this works as you intended.
look at this example:
```
In [1]: s = "aaabbbccc"

In [6]: if "ddd" in s:
   ...:     print ("Yes")
   ...: 

In [7]: if "ddd" and "bbb" in s:
   ...:     print ("Yes")
   ...: 
Yes
```
this is because ```"ddd" and "bbb"``` returns ```"bbb"```

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,134 @@
...
+                count_with_source += 1
+                if magic_runs in filename:
```
jsitarek, line 62: is magic_runs a single run or a list or runs? if the latter this will not work as expected

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,134 @@
...
+                'MAGIC runs': runs_combined
+            })
+        
+    result = pd.DataFrame(result_table)
+    result['MAGIC runs'] = result['MAGIC runs'].apply(fix_lists_and_convert)
+    return(result)
+
+def existing_files( tel_id, date, source, magic_runs ):
+
+    magic_runs = str(magic_runs)
```
jsitarek, line 42: magic_runs is a single run, not a list of runs, right? then a single name "magic_run" would be better.
In general functions should have a doc string explaining the input and output

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,134 @@
...
+        count_with_source = 0  
+        count_with_run_id = 0
+            # Counter for files that include the source.  
+            # Counter for files that include the run_id.
+        for filename in files:
+            if date and source and Y in filename:
+                count_with_source += 1
+                if magic_runs in filename:
+                    count_with_run_id += 1
+        if count_with_source != 0 and count_with_run_id != 0:
```
jsitarek, line 64: count_with_run_id is increased inside an if that increases count_with_sources, so you do not have to check if both are larger then 0, just the one with run

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,134 @@
...
+            print(f"{date}\t{source}\t{magic_runs}\t{count_with_run_id}")
+                    
+def missing_files( tel_id, date, source, magic_runs ):
+    
+    for runs in magic_runs:
+        run = str(runs)
```
jsitarek, line 70: why do you make this look here?
it only keeps the last run from the whole list

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,134 @@
...
+    path = f"/fefs/onsite/common/MAGIC/data/M{tel_id}/event/Calibrated/{YYYY}/{MM}/{DD}" 
+    
+    if os.path.exists(path):
+        files = os.listdir(path)
+        count_with_source = 0  
+        count_with_run_id = 0
+        # Counter for files that include the source. We want to check if any file with the source was found. 
+        # Counter for files that include the run_id. We want to check if any file with the run_id was found. 
+        for filename in files:
+            if date and source and Y in filename:
```
jsitarek, line 89: same problem as above

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,134 @@
...
+                count_with_source += 1
+                for runs in magic_runs:
+                   # run = str(runs)
```
jsitarek, line 92: here you do something strange, you loop over all the runs, but every time you check only the last one (see my comment to line 70)

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
```diff
@@ -0,0 +1,134 @@
...
+            if(tel_id == 1):
+                #Between 2022/09/04 - 2022/12/14 MAGIC 1 had a failure. Therefore we have to skip the range when we want to get information about missing files.
+                if(date<='20220904' or date>='20221214'):
+                    print(f"No files found containing the source '{source}' on {date}")
+                else:
+                    print(f"M1 failure. No files found containing the source '{source}' on {date}.")
+            if(tel_id == 2):
+                print(f"No files found containing the source '{source}' on {date}")
+        if count_with_source != 0 and count_with_run_id == 0:
+            if(date<'20220904' or date>'20221214'):
```
jsitarek, line 105: this should be only for M1
in general I think it would be good to make a list of triples: telescope, date_min, date_max and loop over this list to see if the error is in one of the known bad regions

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/check_MAGIC_runs.py
joannaWojS: Yes, I see... Thanks! I fixed it and I will add all the corrections soon

jsitarek (approved):

Elisa-Visentin:
I just merged the 'base' branch here: this should fix the tests...

PR #222, opened 2024-05-16
Fix bug sources

FedericoDevoto757:
Fix an error in the scripts


Elisa-Visentin:
@jsitarek could you check this? I introduced a bug in the 'management' of the name of the sources, sorry!

jsitarek:
@FedericoDevoto757, can you clean up the PR such that it only has the intended code changes inside? The basic fix is the source name bugfix, and this exist_ok is also harmless, but the changes in LST run list text file file name and Merged directory can be potentially disturbing the code (unless they were intended).
Afterwards we can merge it

FedericoDevoto757:
Good morning, both of the changes to the directory path name and the list_LST file were intended.

jsitarek:
> Good morning, both of the changes to the directory path name and the list_LST file were intended.

thanks, I merged it now

Issue #224, opened 2024-05-23
combo_type a NaN/inf


jsitarek:
I got an error while processing one of LST subruns: dl1_LST-1.Run15757.0043.h5

```
Average offset: 118.780 us
--> Number of coincident events: 1019
--> Fraction over the MAGIC-II events: 62.7%

In total 1566 stereo events are found:
Traceback (most recent call last):
  File "/home/julian.sitarek/anaconda3/envs/magic-lst_0_19/bin/lst1_magic_event_coincidence", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/fefs/home/julian.sitarek/prog/mcp_ctapipe019/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py", line 635, in main
    event_coincidence(
  File "/fefs/home/julian.sitarek/prog/mcp_ctapipe019/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py", line 535, in event_coincidence
    event_data = get_stereo_events(event_data, config)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/fefs/home/julian.sitarek/prog/mcp_ctapipe019/magicctapipe/io/io.py", line 415, in get_stereo_events
    event_data_stereo = event_data_stereo.astype({"combo_type": int})
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/julian.sitarek/anaconda3/envs/magic-lst_0_19/lib/python3.11/site-packages/pandas/core/generic.py", line 6617, in astype
    res_col = col.astype(dtype=cdt, copy=copy, errors=errors)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/julian.sitarek/anaconda3/envs/magic-lst_0_19/lib/python3.11/site-packages/pandas/core/generic.py", line 6640, in astype
    new_data = self._mgr.astype(dtype=dtype, copy=copy, errors=errors)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/julian.sitarek/anaconda3/envs/magic-lst_0_19/lib/python3.11/site-packages/pandas/core/internals/managers.py", line 430, in astype
    return self.apply(
[...]

PR #226, opened 2024-05-23
fix for #224

jsitarek:
the problem was that occasionally two LST-1 images could be assigned to the same MAGIC event. Now in case of such ambiguity both are removed. If there are too many such events >5 the program will exit with error code 11


Elisa-Visentin:
Seems fine. I'm just waiting for the tests

jsitarek:
merging into master.
It might be also useful to merge it into Torino_auto_MCP branch

PR #232, opened 2024-06-11
Possibility to store images 

jsitarek:
(up to stereo reconstruction level)

### DIFF
--- a/magicctapipe/scripts/lst1_magic/config.yaml
+++ b/magicctapipe/scripts/lst1_magic/config.yaml
@@ -63,13 +63,12 @@ MAGIC:
         find_hotpixels: true
         pedestal_type: "from_extractor_rndm"
 
-    save_images: false
-
     muon_ring:
         thr_low: 25
         tailcut: [12, 8]
         ring_completeness_threshold: 25
 
+save_images: false
 
 event_coincidence:
     # History of the time offset between LST-1 and MAGIC
--- a/magicctapipe/scripts/lst1_magic/config.yaml
+++ b/magicctapipe/scripts/lst1_magic/config.yaml
@@ -63,6 +63,8 @@ MAGIC:
         find_hotpixels: true
         pedestal_type: "from_extractor_rndm"
 
+    save_images: false
+
     muon_ring:
         thr_low: 25
         tailcut: [12, 8]
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
@@ -56,6 +56,7 @@ If you want to analyse a target, this is the way to go. See this other script fo
 """
 
 import argparse
+import glob
 import logging
 import sys
 import time
@@ -66,7 +67,11 @@ import numpy as np
 import pandas as pd
 import yaml
 from astropy import units as u
+from astropy.table import vstack
 from ctapipe.instrument import SubarrayDescription
+from ctapipe.io import read_table, write_table
+from lstchain.io import HDF5_ZSTD_FILTERS
+from lstchain.io.io import dl1_images_lstcam_key
 
 from magicctapipe.io import (
     check_input_list,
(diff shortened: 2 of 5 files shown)


gabemery:
@jsitarek 
I am thinking about space usage and wondering if there would be a way to avoid duplicating all images in telescope DL1, coincidence files and stereo files. For the separate telescope to coincidence case I don"t see a simple solution, but is there a reason for having separate coincidence and stereo files?

magicctapipe/scripts/lst1_magic/lst1_magic_stereo_reco.py
```diff
@@ -377,6 +382,30 @@ def stereo_reconstruction(input_file, output_dir, config, magic_only_analysis=Fa
...
+            mask = np.zeros(len(image), dtype=bool)
+            tag = "lst" if tel_id in LSTs_IDs else "magic"
+            obs_evt_ids = event_data.query(f"tel_id=={tel_id}")[
+                [f"obs_id_{tag}", f"event_id_{tag}"]
+            ].to_numpy()
+            for obs_id, evt_id in obs_evt_ids:
+                this_mask = np.logical_and(
+                    image["obs_id"] == obs_id, image["event_id"] == evt_id
+                )
+                mask = np.logical_or(mask, this_mask)
```
gabemery, line 399: It feels like there should be a better way than to loop other all obs_id, evt_id. But I don't have a solution yet.

magicctapipe/scripts/lst1_magic/lst1_magic_stereo_reco.py
```diff
@@ -142,6 +144,10 @@ def stereo_reconstruction(input_file, output_dir, config, magic_only_analysis=Fa
         "mc_tel_ids"
     ]  # This variable becomes a dictionary, e.g.: {'LST-1': 1, 'LST-2': 0, 'LST-3': 0, 'LST-4': 0, 'MAGIC-I': 2, 'MAGIC-II': 3}
 
+    save_images = False
+    if "save_images" in config["MAGIC"]:
+        save_images = config["MAGIC"]["save_images"]
+
```
gabemery, line 150: Can be replaced by :
`save_images = config["MAGIC"].get("save_images", False)`

magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py
```diff
@@ -220,6 +221,11 @@ def magic_calib_to_dl1(
             "MAGIC-Array", tel_positions_magic, tel_descriptions_magic
         )
 
+    save_images = False
+    if "save_images" in config["MAGIC"]:
+        save_images = config["MAGIC"]["save_images"]
```
gabemery, line 226: Can be replaced by :
`save_images = config["MAGIC"].get("save_images", False)`

gabemery (commented):
Mostly good. I have some suggestions.

magicctapipe/scripts/lst1_magic/config.yaml
```diff
@@ -63,6 +63,8 @@ MAGIC:
         find_hotpixels: true
         pedestal_type: "from_extractor_rndm"
 
+    save_images: false
+
```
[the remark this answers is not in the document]
jsitarek: done

magicctapipe/scripts/lst1_magic/lst1_magic_event_coincidence.py
```diff
@@ -157,13 +162,19 @@ def event_coincidence(input_file_lst, input_dir_magic, output_dir, config):
 
     config_coinc = config["event_coincidence"]
 
+    save_images = False
+    if "save_images" in config["MAGIC"]:
+        save_images = config["MAGIC"]["save_images"]
+
```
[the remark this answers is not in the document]
jsitarek: thx, that simplifies the code

magicctapipe/scripts/lst1_magic/lst1_magic_stereo_reco.py
jsitarek: I know what you mean, but I was also not able to find a better way for it

jsitarek:
thank you @gabemery for the review. I implemented your comments.
About the space, the images need to be propagated throughout the whole chain of the file, this is why they end up in all the steps on the way.
But this option would not be enabled by default, and if one processes the files this way, can always remove the intermediate files after they are not needed anymore. Combining coincidence and stereoscopy is a more global issue than this PR. They main reason for the current scheme is that coincidence macro is only run for the data, while stereo reconstruction you need to run both for the data and for MCs.

gabemery (approved):

PR #236, opened 2024-07-05
Tune magic nsb

jsitarek:
a script to derive the missing NSB noise in MAGIC part of the simulations and the modifications to the calibration to allow adding the noise also for MAGIC

### DIFF
--- a/magicctapipe/scripts/lst1_magic/tune_magic_nsb.py
+++ b/magicctapipe/scripts/lst1_magic/tune_magic_nsb.py
@@ -33,41 +33,6 @@ from traitlets.config import Config
 
 log = logging.getLogger(__name__)
 
-parser = argparse.ArgumentParser(description="Tune MAGIC NSB")
-
-# Required arguments
-parser.add_argument(
-    "--config",
-    "-c",
-    type=Path,
-    help="Path to the configuration file for the production (MCP general config)",
-    required=True,
-)
-
-parser.add_argument(
-    "--input-mc",
-    "-m",
-    type=Path,
-    help="Path to a simtel file of the production (must include the true "
-    "p.e. images)",
-    required=True,
-)
-
-parser.add_argument(
-    "--input-data",
-    "-d",
-    type=Path,
-    help="Path to a data _Y_ MARS file",
-    required=True,
-)
-
-parser.add_argument(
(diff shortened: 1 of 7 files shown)


Elisa-Visentin:
Hi @jsitarek, I was checking the PR. Just a naïve question: why do we need two different NSB corrections for the 2 MAGIC? It is due to a different level of simulated (DL0) background, to the different features of the telescopes or to the (slightly)  different FOV?

jsitarek:
Hi @Elisa-Visentin 
The code is done LST-style, i.e. you make a distribution of noise in MC and in data and using Gaussian approximations you add the difference. When you do the same procedure for M1 and M2 you would always get somewhat different results. You are right that in principle the difference should be small, but the two telescopes are not the same, in particular the afterpulse distribution in PMTs is different for M1 and M2, plus the simulation of electronic noise and NSB has some simplifications. From the tests that I was running I saw that the extra_noise parameters, both for dim and bright pixels end up very similar for M1 and M2, but there is a difference in added bias, it is very small for M2, but there is some bias to be added in M1.

Elisa-Visentin:
Ok thank you!

magicctapipe/scripts/lst1_magic/tune_magic_nsb.py
```diff
@@ -0,0 +1,334 @@
...
+import yaml
+from ctapipe.calib.camera import CameraCalibrator
+from ctapipe.io import EventSource
+from ctapipe_io_magic import MAGICEventSource
+from scipy.interpolate import interp1d
+from traitlets.config import Config
+
+log = logging.getLogger(__name__)
+
+parser = argparse.ArgumentParser(description="Tune MAGIC NSB")
```
Elisa-Visentin, line 36: Maybe the parser could be put in the main (instead of using global variables), but not a real issue (like this it is fine and it runs)

Elisa-Visentin (approved):

Elisa-Visentin:
There seems to be a failing test due to no M1 M2 stereo events in the training sample. I will check it on my laptop asap just to be sure that everything is fine (i.e., only statistics)

Elisa-Visentin (commented):
Tests to be fixed before merging

PR #240, opened 2024-07-29
Auto update MAGIC database

joannaWojS:
Added new script (update_MAGIC_database.py) for automatically updating the MAGIC database.

### DIFF
--- a/magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
+++ b/magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
@@ -254,10 +254,10 @@ def create_new_database(df, date_min, date_max, tel_id):
             if count_with_run_id != 0:
                 new_rows.append(
                     {
-                        "DATE": date,
-                        "source": source,
-                        "Run ID": run_id,
-                        "number of subruns": count_with_run_id,
+                        "DATE": int(date),
+                        "source": str(source),
+                        "Run ID": int(run_id),
+                        "number of subruns": int(count_with_run_id),
                     }
                 )
 
--- a/magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
+++ b/magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
@@ -1,8 +1,7 @@
 """
 The script updates the common MAGIC database from a given time range.
 At the moment, to avoid accidentally destroying the previous database,
-we save the updated database as a new file (see in main function new_h5_file_path
-and new_database_file_path).
+we save the updated database as a new file (see in main function new_h5_file_path).
 If the path to the previous database is not found,
 the script creates a new one. The start of the time interval
 is the date of the beginning of the common MAGIC+LST1 observations.
@@ -17,15 +16,14 @@ In this path, 'tel_id' refers to the telescope ID, which must be either 1 or 2.
 """
 
 import argparse
-import sys
 import os
 from datetime import datetime, timedelta
 
+import numpy as np
 import pandas as pd
 import yaml
 
 from magicctapipe.io import resource_file
-import numpy as np
 
 
 def fix_lists_and_convert(cell):
@@ -73,7 +71,7 @@ def table_magic_runs(df, date_min, date_max):
     result_table = []
 
     for (date, source), group in grouped_data:
-        if date >= date_min and date <= date_max: 
+        if date >= date_min and date <= date_max:
             runs_combined = group["MAGIC_runs"].sum()
 
             result_table.append(
@@ -134,7 +132,9 @@ def update_tables(database, DF, tel_id):
         non_matching_rows_reset = non_matching_rows.reset_index(drop=True)
         new_rows = []
 
-        for _, (date, source, run_id) in non_matching_rows_reset[["DATE", "source", "Run ID"]].iterrows(): 
+        for _, (date, source, run_id) in non_matching_rows_reset[
+            ["DATE", "source", "Run ID"]
+        ].iterrows():
 
             run_id = str(run_id)
             date_obj = datetime.strptime(date, "%Y%m%d")
@@ -180,7 +180,9 @@ def update_tables(database, DF, tel_id):
         combined_df.reset_index(drop=True, inplace=True)
 
         for column in combined_df.columns[4:]:
-            combined_df[column] = combined_df[column].replace(r'^\s*$', np.nan, regex=True)
+            combined_df[column] = combined_df[column].replace(
+                r"^\s*$", np.nan, regex=True
+            )      
             not_null_data = combined_df[column].dropna()
             if not_null_data.empty:
                 continue  # Skip if all values are NaN
@@ -208,12 +210,13 @@ def update_tables(database, DF, tel_id):
 
     return combined_df
 
+
 def main():
 
     """Main function."""
-    
+
     parser = argparse.ArgumentParser()
-    
+
     date_min_default = "20191101"
     current_datetime = datetime.now()
     date_max_default = current_datetime.strftime("%Y%m%d")
@@ -224,7 +227,7 @@ def main():
         dest="date_min",
         type=str,
         default=date_min_default,
-        help="Start of the time interval (in LST convention, format YYYYMMDD)."
+        help="Start of the time interval (in LST convention, format YYYYMMDD).",
     )
 
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
```diff
@@ -0,0 +1,351 @@
...
+
+    """
+    Generate a table with data filtered by the specified date range.
+
+    Parameters
+    ----------
+    df : pandas.DataFrame
+        DataFrame with general information about MAGIC+LST1 observations.
+    date_min : str
+        Start of the time interval (in LST convention).
```
jsitarek, line 53: maybe give example, since this is a string it is not clear if it should be e.g. 2022.10.01 or 20221001 or 2022_10_01

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
```diff
@@ -0,0 +1,351 @@
...
+    date_max : str
+        End of the time interval (in LST convention).
+
+    Returns
+    -------
+    pandas.DataFrame
+        A DataFrame filtered by the specified date range.
+    """
+
+    df_selected_data = df.iloc[:, [2, 1, 25]]
```
jsitarek, line 63: here you should check if you really have date, source and MAGIC_runs at the positions 2, 1 and 25.
or alternatively access columns by names

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
```diff
@@ -0,0 +1,351 @@
...
+    df_selected_data.columns = ["DATE", "source", "MAGIC_runs"]
+    grouped_data = df_selected_data.groupby(["DATE", "source"])
+    result_table = []
+
+    for (date, source), group in grouped_data:
+        if date >= date_min and date <= date_max: 
+            runs_combined = group["MAGIC_runs"].sum()
```
jsitarek, line 70: did you check if this does not add the same MAGIC run multiple times if it is mentioned in different LST runs ?

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
```diff
@@ -0,0 +1,351 @@
...
+
+def update_tables(database, DF, tel_id):
+    """
+    Updating the MAGIC database by comparison of data that are only in
+    common MAGIC+LST1 database and not in the MAGIC database.
+    Then, the function checks existing files and counts number of subruns.
+    Data are added chronologically.
+
+    The updated table DF may include new rows that contain NaN values in some cells.
+    The function automatically filling NaN values withpredefined default values
```
jsitarek, line 91: space missing

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
```diff
@@ -0,0 +1,351 @@
...
+        raise Exception("There is no un-updated data for a given time interval. ")
+    else:
+
+        non_matching_rows_reset = non_matching_rows.reset_index(drop=True)
+        new_rows = []
+
+        for index, row in non_matching_rows_reset.iterrows():
+            date = row["DATE"]
+            source = row["source"]
+            run_id = row["Run ID"]
```
jsitarek, line 133: you can combine this and next line

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
```diff
@@ -0,0 +1,351 @@
...
+        new_rows["DATE"] = pd.to_datetime(new_rows["DATE"])
+        combined_df = pd.concat([DF, new_rows], ignore_index=True)
+        combined_df["DATE"] = pd.to_datetime(combined_df["DATE"], errors="coerce")
+        combined_df = combined_df.sort_values("DATE")
+
+        combined_df["DATE"] = combined_df["DATE"].dt.strftime("%Y%m%d")
+        combined_df["Run ID"] = combined_df["Run ID"].astype(int)
+        combined_df.reset_index(drop=True, inplace=True)
+
+        for column in combined_df.columns[4:]:
```
jsitarek, line 176: this also loops over the "old" lines in the table, and some of those might not be processed yet, did you check if there are no problems caused by this?

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
```diff
@@ -0,0 +1,351 @@
...
+                combined_df[column]
+                .fillna(default_value)
+                .astype(type(not_null_data.iloc[0]))
+            )
+
+        combined_df = combined_df.infer_objects()
+
+    return combined_df
+
+def create_new_database(df, date_min, date_max, tel_id):
```
jsitarek, line 204: this function has a lot of common code with the previous one
This is why I was suggesting to have only one function that has previous table by default as Null, and if so it first creates a new empty table, and then runs updating function on this empty table

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
```diff
@@ -0,0 +1,351 @@
...
+
+        print("Updating database...")
+
+        database = table_magic_runs(df, date_min, date_max)
+
+        for tel in tel_id:
+
+            DF = pd.read_hdf(
+                previous_database_path,
+                key=f"MAGIC{tel}/runs_M{tel}",
```
jsitarek, line 297: the same key you will use also for writing so you can put it into a variable

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
```diff
@@ -0,0 +1,351 @@
...
+        key="str/table",
+    )
+
+    # Set "" to generate a new database.
+    previous_database_path = "/fefs/aswg/workspace/joanna.wojtowicz/data/Common_MAGIC_LST1_data_MAGIC_runs_subruns.h5"
+    file_exists = os.path.exists(previous_database_path)
+
+    if file_exists:
+
+        # TO DO : set time interval - format YYYYMMDD
```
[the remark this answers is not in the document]
joannaWojS: I was wondering about it, but I didn’t have enough time to change it before my vacation. Anyway, I think it’s a good idea and will change it soon ;)

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
joannaWojS: Yes, this loop iterates over each cell/line in a column. The aim of this part of the code is to avoid discrepancies in data types before saving a file as HDF5. I assumed that if a file has not been processed yet, the default value in a cell would be NaN. Then, lines with NaN are skipped, and we look for a cell with a value in the column to check its type. This type is then used to define the rest of the column with a new default value. I hope that the default value used is neutral enough. If the entire column contains NaN (indicating that no files have been processed yet), we skip the column, avoiding any issues with data types. However, after your question, I realized that a cell could contain an empty string or only whitespace. I will take this into account when updating the code.

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
joannaWojS: I see... And I think I just did it :) New version of the script will appear soon.

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
```diff
@@ -0,0 +1,322 @@
...
+        print("Database does not exist. Creating a new database...")
+        
+        new_database_file_path = "/fefs/aswg/workspace/joanna.wojtowicz/data/Common_MAGIC_LST1_data_MAGIC_runs_subruns_NEW.h5"
+
+        database_default = table_magic_runs(df, date_min_default, date_max_default)
+
+        for tel in tel_id:
+            
+            tel_roman = roman_numerals[tel]
+            key = config_dict["database_keys"][f"MAGIC-{tel_roman}"]
```
jsitarek, line 306: in other parts of the code if tel_id was 1 or 2 for MAGIC-I or MAGIC-II I was using:
```"MAGIC-" + tel_id * "I"```
your version is more direct but have some extra lines of code. You can either keep it as it is or change it, as you prefer

File: magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py [hunk 11]
```diff
@@ -0,0 +1,322 @@
...
+
+            DF = pd.read_hdf(
+                previous_database_path,
+                key=key,
+            )
+
+            updated_df = update_tables(database, DF, tel)
+            print(updated_df)
+
+        try:
```
jsitarek, line 289: you can extract this try except part after the if because saving is the same for both if cases, you just need to use the same variable for output path

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
joannaWojS: I understand, but without this step, I am unable to save my tables to an HDF5 file due to mixed data types. For example, I encountered the following error: 'An error occurred: Cannot serialize the column [DL1_rc_all] because its data contents are not [string] but [mixed] object dtype.

joannaWojS: After resolving this issue, we should be able to merge the branches, as the current version has no linting, pyflakes, or docs warnings.

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/update_MAGIC_database.py
jsitarek: you marked it with thumb up, but the code still has two separate writing blocks. 
But it is not a big problem, we can merge it as it is and prepare a refactoring PR later on

PR #241, opened 2024-08-07
Script to apply cloud affection correction to the LST, Magic 1, and M…

nzywucka:
…agic 2 data


magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+from magicctapipe.io import save_pandas_data_in_table
+
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.StreamHandler())
+logger.setLevel(logging.INFO)
+
+
+def model0(imp, h, zd):
+    """
+    Calculates distance
```
jsitarek, line 40: Calculated the geometrical part of the model relating the emission height with the angular distance from the arrival direction

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+
+    Parameters
+    ----------
+    imp : float
+        Impact
+    h : float
+        Height
+    zd : float
+        Zenith distance
```
jsitarek, line 49: if the values are floats you should specify units

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+    float
+        Angular distance in units of degree
+    """
+    d = h / np.cos(zd)
+    return np.arctan((imp / d).to("")).to_value("deg")
+
+
+def model2(imp, h, zd):
+    """
+    Calculates model
```
jsitarek, line 62: Calculates the phenomenological correction to the distances obtained with model0

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+
+    Parameters
+    ----------
+    imp : float
+        Impact
+    h : float
```
jsitarek, line 68: you add h to H0 and latter one is Quantity to h cannot be float.
once you correct it, note also that model0 and model2 should have the same descriptions of parameters

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+        Height
+    zd : float
+        Zenith distance
+
+    Returns
+    -------
+    float
+        Bias
```
jsitarek, line 76: it does not return a bias, but the distance like in model0, just corrected for bias

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+def model0(imp, h, zd):
+    """
+    Calculates distance
+
+    Parameters
+    ----------
+    imp : float
+        Impact
+    h : float
+        Height
```
jsitarek, line 47: you should specify if it is a.s.l. or a.g.l.

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+        Bias
+    """
+    H0 = 2.2e3 * u.m
+    bias = 0.877 + 0.015 * ((h + H0) / (7.0e3 * u.m))
+    return bias * model0(imp, h, zd)
+
+
+def trans_height(x, Hc, dHc, trans):
+    """
+    Calculates height of a cloud transmission
```
jsitarek, line 85: Calculates transmission from a geometrically broad cloud at a set of heights.
Cloud is assumed to be homegeneous

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+
+    Parameters
+    ----------
+    x : float
+        Position of each layer
```
jsitarek, line 90: height of each layer 
a.s.l./a.g.l.?
units

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+    Hc : float
+        Cloud above the telescope
```
jsitarek, line 92: Height of the base of the cloud
a.s.l.? a.g.l.?

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+    dHc : float
+        Cloud thickness
+    trans : float
+        Transmission of the cloud
+
+    Returns
+    -------
+    float
+        Height of a cloud transmission
```
jsitarek, line 101: the function returns transmissions not height

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+    input_file, output_file, telid, focal_eq, focal_eff, camgeom
+):
+    """
+    Corrects LST-1 and MAGIC data affected by a cloud presence
+
+    Parameters
+    ----------
+    input_file : str
+        Path to an input .h5 DL1 file
+    output_file : str
```
jsitarek, line 119: you do not use output_file in this function

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+        Path to a directory where to save an output corrected DL1 data file
+    telid : int
+        LST-1 and MAGIC telescope ids
+    focal_eq : float
+        Equivalent focal length
+    focal_eff : float
+        Effective focal length
```
jsitarek, line 126: what is the difference between effective and equivalent focal length?
either way you use only one in this function

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+    camgeom : int
```
jsitarek, line 127: camera geometry is not an int

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+        Camera geometry
+
+    Returns
+    -------
+    pandas.core.frame.DataFrame
+        Data frame of corrected DL1 parameters
+    """
+
+    Hc = 5900 * u.m  # cloud above the telescope
```
jsitarek, line 136: ultimately the cloud profile should be read from LIDAR, for now we can introduce an option to read it from the config file something like
```cloud_image_correction:
  specify_cloud: yes
  base_height: "... m"
  thickness: "... m"
  vertical_transmission: ....```

File: magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py [hunk 15]
```diff
@@ -0,0 +1,417 @@
...
+        obs_id_lst = dl1_params["obs_id_lst"][index]
+        event_id = dl1_params["event_id_magic"][index]
+        obs_id = dl1_params["obs_id_magic"][index]
+        if telid == 1:
+            event_id, obs_id = event_id_lst, obs_id_lst
+
+        pointing_az = dl1_params["pointing_az"][index]
+        pointing_alt = dl1_params["pointing_alt"][index]
+        pointing_alt_deg = np.rad2deg(pointing_alt)
+        zenith = 90.0 - pointing_alt_deg
```
jsitarek, line 157: since you do not use pointing_alt_deg anywhere else I could combine this and previous line

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+        log_intensity = np.log10(dl1_params["intensity"])[index]
+        timestamp = dl1_params["timestamp"][index]
+        time_diff = dl1_params["time_diff"][index]
+        n_islands = dl1_params["n_islands"][index]
+        n_pixels = dl1_params["n_pixels"][index]
+        signal_pixels = dl1_params["n_pixels"][index]
+
+        trans = 0.58  # transmission of the cloud
+        trans = trans ** (1 / np.cos(zenith))  # .to_value("")
+        nlayers = 10  # to simplify calculations with the model cloud is considered to be composed of multiple layers
```
jsitarek, line 172: this can be also read from config file

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+        Hcl = np.linspace(Hc, Hc + dHc, nlayers)  # position of each layer
+        transl = trans_height(Hcl, Hc, dHc, trans)  # transmissions of each layer
+        transl = np.append(
+            transl, transl[-1]
+        )  # add one more element for heights above the top of the cloud
+
+        alt_rad = np.deg2rad(dl1_params["alt"][index])
+        az_rad = np.deg2rad(dl1_params["az"][index])
+
+        eff_focal = focal_eq[telid]  # *u.m # m
```
jsitarek, line 182: in docstring you wrote that it is a float, but here you use it as an array or dictionary

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+        m2deg = np.rad2deg(1) / eff_focal * u.degree
```
jsitarek, line 183: this and the previous line can go before the event loop

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+        signal_pixels = dl1_params["n_pixels"][index]
+
+        trans = 0.58  # transmission of the cloud
+        trans = trans ** (1 / np.cos(zenith))  # .to_value("")
+        nlayers = 10  # to simplify calculations with the model cloud is considered to be composed of multiple layers
+        Hcl = np.linspace(Hc, Hc + dHc, nlayers)  # position of each layer
+        transl = trans_height(Hcl, Hc, dHc, trans)  # transmissions of each layer
+        transl = np.append(
+            transl, transl[-1]
+        )  # add one more element for heights above the top of the cloud
```
jsitarek, line 177: all those calculations of the layers of the cloud are generic and can you do it once before events loop to save time

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+        )
+
+        tel_frame = TelescopeFrame(telescope_pointing=telescope_pointing)
+        tel = pointing_altaz.transform_to(tel_frame)
+
+        src_x = tel.fov_lat.to(u.deg).value
+        src_y = tel.fov_lon.to(u.deg).value
+
+        # Transform to Engineering camera
+        src_x, src_y = -src_y * u.deg, -src_x * u.deg
```
jsitarek, line 205: you just dropped deg and here you multiply by it again

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+        ilayer = np.digitize(distance, dist_corr_layer)
+        trans_pixels = transl[ilayer]
+
+        inds_img = np.where(
+            (dl1_images["event_id"] == event_id)
+            & (dl1_images["tel_id"] == telid)
+            & (dl1_images["obs_id"] == obs_id)
+        )[0]
+
+        if len(inds_img) > 0:
```
jsitarek, line 231: this condition should always be true if we have images in the files. 
you should add at least some error message if you find an event without image rather then skipping it silently

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+            for index_img in inds_img:
```
jsitarek, line 232: why the for loop?
it should be always exactly one image?
finding more then 1 image corresponding to a given set of parameters is also an error (like not finding any image)

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+
+                source_pos_in_camera = sky_to_camera(
+                    alt_rad * u.rad,
+                    az_rad * u.rad,
+                    focal_eq[telid],
+                    pointing_alt * u.rad,
+                    pointing_az * u.rad,
+                )
+
+                disp_parameters = disp.disp(
```
jsitarek, line 259: did you check if the disp_parameters are exactly the same as in the original file if you specify T=1 cloud? We need to make sure that proper focal length is used 5 lines up

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+                    x, y, source_pos_in_camera.x, source_pos_in_camera.y, psi
+                )
+
+                disp_dx = disp_parameters[0].value * u.m
+                disp_dy = disp_parameters[1].value * u.m
+                disp_norm = disp_parameters[telid].value * u.m
```
jsitarek, line 265: something looks wrong here, first you treat it as an array with different parameters at different indicies, and here you treat it as an array with indices corresponding to telid

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+                        "intercept": timing_params.intercept,
+                        "deviation": timing_params.deviation,
+                        "intensity_width_1": leakage_params.intensity_width_1,
+                        "intensity_width_2": leakage_params.intensity_width_2,
+                        "pixels_width_1": leakage_params.pixels_width_1,
+                        "pixels_width_2": leakage_params.pixels_width_2,
+                        "conc_cog": conc_params.cog,
+                        "conc_core": conc_params.core,
+                        "conc_pixel": conc_params.pixel,
+                        "impact": impact,
```
jsitarek, line 298: you should not save parameters like impact, alt, az, etc., because they were not recomputed after the correction. 
The file that you save should have DL1-like format and then we need to run DL1stereo macro over it again to get corrected stereoscopic reconstruction

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+                        "pointing_alt": pointing_alt,
+                        "pointing_az": pointing_az,
+                        "alt": alt_rad,
+                        "az": az_rad,
+                        "wl": wl,
```
jsitarek, line 303: you save here wl from before correction

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+
+                disp_dx = disp_parameters[0].value * u.m
+                disp_dy = disp_parameters[1].value * u.m
+                disp_norm = disp_parameters[telid].value * u.m
+                disp_angle = disp_parameters[3].value * u.rad
+                disp_sign = disp_parameters[4]
+                src_x = source_pos_in_camera.x.value * u.m
+                src_y = source_pos_in_camera.y.value * u.m
+
+                data.append(
```
jsitarek, line 271: as we were discussing specifing all the parameters by hand makes the code unnecessarily complicated, much harder to maintain and prone to errors (there are some below)
you should save the whole containers, here you have an example how to do it:
https://github.com/cta-observatory/magic-cta-pipe/blob/352c56fdf11237322c41f055f0c976f4b34ef8fb/magicctapipe/scripts/lst1_magic/magic_calib_to_dl1.py#L394-L397

you will just need to stack them up from all the telescopes

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
jsitarek, line 259: actually, why do you recalculate disp? I do not think it is there in the DL1 files.

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+
+    parser.add_argument(
+        "--output_file",
+        "-o",
+        dest="output_file",
+        type=str,
+        default="./data",
+        help="Path to a directory where to save an output corrected DL1 file",
+    )
+
```
jsitarek, line 345: you should also parse the path to the config file

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+    args = parser.parse_args()
+
+    subarray_info = SubarrayDescription.from_hdf(args.input_file)
+
+    tel_descriptions = subarray_info.tel
+    camgeom = {}
+    for telid, telescope in tel_descriptions.items():
+        camgeom[telid] = telescope.camera.geometry  # 1 - LSTCam; 2,3 - MAGICCam
```
jsitarek, line 353: the comment: "# 1 - LSTCam; 2,3 - MAGICCam" is only valid for MAGIC+LST1 data

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+    focal_eq = {}
+    focal_eff = {}
+
+    for telid, telescope in tel_descriptions.items():
+        optics_row = optics_table[optics_table["optics_name"] == telescope.name]
+        if len(optics_row) > 0:
+            focal_length_eq = optics_row["equivalent_focal_length"][0]
+            focal_eq[telid] = focal_length_eq * u.m
+            focal_length_eff = optics_row["effective_focal_length"][0]
+            focal_eff[telid] = focal_length_eff * u.m
```
jsitarek, line 367: I would do multiplication by u.m immediately, and also I think you only need one, right?

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+
+    df_lst = process_telescope_data(
```
jsitarek, line 369: why not a loop over all the telescopes in the file?
and you can extract which telescopes to use from the config file as well

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+    df_m1 = process_telescope_data(
+        args.input_file, args.output_file, 2, focal_eq, focal_eff, camgeom
+    )
+    df_m2 = process_telescope_data(
+        args.input_file, args.output_file, 3, focal_eq, focal_eff, camgeom
+    )
+
+    df_all = pd.concat([df_lst, df_m1, df_m2], ignore_index=True)
+
+    columns_to_convert = [
```
jsitarek, line 381: all this code with the conversion also will not be needed if you write the way how I specified above

jsitarek:
Hi @nzywucka 
can you post the .keys() from the /events/parameters from original DL1 file (not the DL1stereo!) and from this new file that you create?

also please check that all the parameters have exactly the same values if T=1 is taken

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Changed

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: changed

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Changed

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Done

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Done

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Done

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Done

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Done

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: done

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: changed

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Done

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: I'm reading calculating zenith for each event with pointing alt parameter from the table:
pointing_alt = dl1_params['pointing_alt'][index]
zenith = 90.0 - np.rad2deg(pointing_alt)
so this part needs to be in the loop.

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Changed

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,417 @@
...
+        eff_focal = focal_eq[telid]  # *u.m # m
+        m2deg = np.rad2deg(1) / eff_focal * u.degree
+
+        impact = dl1_params["impact"][index] * u.m
+        psi = dl1_params["psi"][index] * u.deg
+        cog_x = (dl1_params["x"][index] * m2deg).value * u.deg
+        cog_y = (dl1_params["y"][index] * m2deg).value * u.deg
+
+        # Source position
+        pointing_altaz = SkyCoord(alt=alt_rad * u.rad, az=az_rad * u.rad, frame=AltAz())
```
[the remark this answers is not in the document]
nzywucka: Right

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Removed from the saving part.

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Right, changed that as well as log_intensity

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: I removed calculation of the disp parameters from the script.

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Now I'm saving the parameters with pandas:

save_pandas_data_in_table(
    df_all, output_file, group_name="/events", table_name="parameters"
)

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: This is anyway removed from the code now.

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: The for loop here is just to use indexing.

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Sure, the following line is added:

else:
    raise ValueError("Error: 'inds_img' list is empty!")

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: This was my working comment - I deleted it.

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: done

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Done

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Right

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: A loop over all the telescopes is added. I will also add info from the config file.

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Done.

magicctapipe/scripts/lst1_magic/config.yaml
```diff
@@ -82,7 +67,7 @@ MAGIC:
         tailcut: [12, 8]
         ring_completeness_threshold: 25
 
-save_images: false
+save_images: true
```
jsitarek, line 70: while this is required for the cloud correction code to work, saving images increases the size of the files considerable, therefore in the default config file the option should be kept as false

magicctapipe/scripts/lst1_magic/config.yaml
```diff
@@ -53,20 +52,6 @@ MAGIC:
         use: true
         factor: 1.143
 
-    increase_nsb_m1:
```
jsitarek, line 56: it seems that you used an older config file and now are trying to merge changes that counteract previous changes

magicctapipe/scripts/lst1_magic/config.yaml
```diff
@@ -279,3 +264,11 @@ dl2_to_dl3:
     source_name: "Crab"
     source_ra: null  # used when the source name cannot be resolved
     source_dec: null  # used when the source name cannot be resolved
+    
+cloud_correction:
+    specify_cloud: yes
+    base_height: 5900
```
jsitarek, line 270: we could either define it as cm in the config file, or add a comment that it is in cm

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+"""
+This script corrects LST-1 and Magic data for the cloud affection.
```
jsitarek, line 5: MAGIC with capital letters

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+
+
+def model0(imp, h, zd):
+    """
+    Calculated the geometrical part of the model relating the emission height with the angular distance from the arrival direction
+
+    Parameters
+    ----------
+    imp : astropy.units.quantity.Quantity
+        Impact in m
```
jsitarek, line 43: if it is a quantity it does not matter if it is in m or other distance units

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+    h : astropy.units.quantity.Quantity
```
jsitarek, line 44: the way it is used in the code it is an array of quantities

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+
+
+def model2(imp, h, zd):
+    """
+    Calculates the phenomenological correction to the distances obtained with model0
+
+    Parameters
+    ----------
+    imp : astropy.units.quantity.Quantity
+        Impact in m
```
jsitarek, line 65: see above

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+    h : astropy.units.quantity.Quantity
```
jsitarek, line 66: see above

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+        Height of each cloud layer a.g.l.
+    zd : numpy.float64
+        Zenith distance in deg
+
+    Returns
+    -------
+    astropy.units.quantity.Quantity
```
jsitarek, line 73: the return type should be the same as in model0

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+
+
+def trans_height(x, Hc, dHc, trans):
+    """
+    Calculates transmission from a geometrically broad cloud at a set of heights.
+    Cloud is assumed to be homegeneous.
+
+    Parameters
+    ----------
+    x : astropy.units.quantity.Quantity
```
jsitarek, line 88: array?

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+    input_file : str
+        Path to an input .h5 DL1 file
+    config : dict
+        Configuration for the LST-1 + MAGIC analysis
+    telid : numpy.int16
+        LST-1 and MAGIC telescope ids
+    focal_eff : dict
+        Effective focal length
+    camgeom : dict
+        An instance of the CameraGeometry class containing information about the
```
jsitarek, line 123: since CameraGeometry is one of the standard classes, I do not think you need to describe it in so much detail. On the other hand, the type is dict, while here you talk about CameraGeometry, so better to say that it is a dictionary of the camera geometries for each tel_id

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+    Parameters
+    ----------
+    input_file : str
+        Path to an input .h5 DL1 file
+    config : dict
+        Configuration for the LST-1 + MAGIC analysis
+    telid : numpy.int16
+        LST-1 and MAGIC telescope ids
+    focal_eff : dict
+        Effective focal length
```
jsitarek, line 121: since the function runs over a single tel_id at the time, why not pass directly the focal length of this telescope instead of the whole dictionary?

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+
+    inds = np.where(
+        np.logical_and(dl1_params["intensity"] > 0.0, dl1_params["tel_id"] == telid)
+    )[0]
+    for index in inds:
+        event_id_lst = dl1_params["event_id_lst"][index]
+        obs_id_lst = dl1_params["obs_id_lst"][index]
+        event_id = dl1_params["event_id_magic"][index]
+        obs_id = dl1_params["obs_id_magic"][index]
+        if telid == 1:
```
jsitarek, line 150: this is assuming that we are using the default naming of the telescopes.

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+            event_id, obs_id = event_id_lst, obs_id_lst
+
+        pointing_az = dl1_params["pointing_az"][index]
+        pointing_alt = dl1_params["pointing_alt"][index]
+        zenith = 90.0 - np.rad2deg(pointing_alt)
+        psi = dl1_params["psi"][index] * u.deg
+        time_diff = dl1_params["time_diff"][index]
+        n_islands = dl1_params["n_islands"][index]
+        n_pixels = dl1_params["n_pixels"][index]
+        signal_pixels = dl1_params["n_pixels"][index]
```
jsitarek, line 160: why do you have 2 variables for the same quantity?

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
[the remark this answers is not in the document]
jsitarek: you wrote "right", but it is still called "pointing" in the code

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+                image = dl1_images["image"][index_img]
+                cleanmask = dl1_images["image_mask"][index_img]
+                peak_time = dl1_images["peak_time"][index_img]
+                image /= trans_pixels
+                corr_image = image.copy()
+                corr_image[~cleanmask] = 0
+
+                hillas_params = hillas_parameters(camgeom[telid], corr_image)
+                timing_params = timing_parameters(
+                    camgeom[telid], corr_image, peak_time, hillas_params
```
jsitarek, line 233: if you would pass directly camgeom[telid] to the function as I suggested above you would not have to access it here in 3 different places

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+                )
+                leakage_params = leakage_parameters(
+                    camgeom[telid], corr_image, signal_pixels
```
jsitarek, line 236: this looks wrong. you pass here n_pixels, while the leakage calculation code requires cleaning mask: https://ctapipe.readthedocs.io/en/v0.19.3/api/ctapipe.image.leakage_parameters.html

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
jsitarek: I do not understand how the current code works. 
you do not add "config" to the parser, but below in line 312 you use it

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
jsitarek: ok, but why you still multiply by u.m in line 310 and not directly in 309 assigning it to focal_eff[telid]?

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+        camgeom[telid] = telescope.camera.geometry
+
+    optics_table = read_table(
+        args.input_file, "/configuration/instrument/telescope/optics"
+    )
+    focal_eff = {}
+
+    for telid, telescope in tel_descriptions.items():
+        optics_row = optics_table[optics_table["optics_name"] == telescope.name]
+        if len(optics_row) > 0:
```
jsitarek, line 308: you really need this focal length for the code to work, so if optics_row is empty you should raise an exception

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: Done.

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+def model0(imp, h, zd):
+    """
+    Calculated the geometrical part of the model relating the emission height with the angular distance from the arrival direction
+
+    Parameters
+    ----------
+    imp : astropy.units.quantity.Quantity
+        Impact in m
+    h : astropy.units.quantity.Quantity
+        Height of each cloud layer a.g.l.
```
jsitarek, line 45: Array with heights of each ...

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+        pointing_alt = dl1_params["pointing_alt"][index]
+        zenith = 90.0 - np.rad2deg(pointing_alt)
+        psi = dl1_params["psi"][index] * u.deg
+        time_diff = dl1_params["time_diff"][index]
+        n_islands = dl1_params["n_islands"][index]
+        n_pixels = dl1_params["n_pixels"][index]
+        signal_pixels = dl1_params["n_pixels"][index]
+
+        Hc = correction_params.get("base_height") * u.m
+        dHc = correction_params.get("thickness") * u.m
```
jsitarek, line 163: move before the for loop

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
...
+                    **timing_params,
+                    **leakage_params,
+                    **conc_params,
+                }
+
+                # Add real event information (assuming it's also a dictionary-like object)
+                event_info_dict = {
+                    "obs_id": obs_id,
+                    "event_id": event_id,
+                    "pointing_alt": pointing_alt,
```
jsitarek, line 253: add tel_id

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,349 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+"""
+This script corrects LST-1 and Magic data for the cloud affection.
+
+Usage:
+$ python lst_m1_m2_cloud_correction.py
+--input-file dl1_stereo/dl1_LST-1_MAGIC.Run03265.0040.h5
+(--output-dir dl1_corrected)
```
jsitarek, line 10: and config file ?

magicctapipe/scripts/lst1_magic/config.yaml
```diff
@@ -279,3 +277,11 @@ dl2_to_dl3:
     source_name: "Crab"
     source_ra: null  # used when the source name cannot be resolved
     source_dec: null  # used when the source name cannot be resolved
+    
+cloud_correction:
```
jsitarek, line 281: the changes to this config file are fine, but the file in the meantime was moved to a different location (resources directory), and I'm not sure if this will get merged properly

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,360 @@
...
+
+    Parameters
+    ----------
+    input_file : str
+        Path to an input .h5 DL1 file
+    config : dict
+        Configuration for the LST-1 + MAGIC analysis
+    tel_id : numpy.int16
+        LST-1 and MAGIC telescope ids
+    camgeom : float
```
jsitarek, line 121: it is not a float

nzywucka: ctapipe.instrument.camera.geometry.CameraGeometry

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
nzywucka: added

magicctapipe/scripts/lst1_magic/config.yaml
nzywucka: I've now done:
git fetch origin
git merge origin/master
to merge the newest version of master into correction branch.

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,361 @@
...
+    dHc = u.Quantity(correction_params.get("thickness"))
+    trans0 = correction_params.get("vertical_transmission")
+    nlayers = correction_params.get("number_of_layers")
+
+    all_params_list = []
+
+    dl1_params = read_table(input_file, "/events/parameters")
+    dl1_images = read_table(input_file, "/events/dl1/image_" + str(tel_id))
+
+    focal = focal_eff[tel_id]
```
jsitarek, line 146: you write above that it is a float, but here you treat it as a dictionary (that you indeed pass to the function in main)

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,361 @@
...
+
+    inds = np.where(
+        np.logical_and(dl1_params["intensity"] > 0.0, dl1_params["tel_id"] == tel_id)
+    )[0]
+    for index in inds:
+        event_id_lst = dl1_params["event_id_lst"][index]
+        obs_id_lst = dl1_params["obs_id_lst"][index]
+        event_id = dl1_params["event_id_magic"][index]
+        obs_id = dl1_params["obs_id_magic"][index]
+        if tel_id == 1:
```
jsitarek, line 157: you marked this as resolved, but the code is still the same, you pass config to the function so you can extract mc_tel_id from it, and check if tel_id is equal to the one written in mc_tel_id for LST-1

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -0,0 +1,362 @@
...
+
+    Parameters
+    ----------
+    input_file : str
+        Path to an input .h5 DL1 file
+    config : dict
+        Configuration for the LST-1 + MAGIC analysis
+    tel_id : numpy.int16
+        LST-1 and MAGIC telescope ids
+    camgeom : str
```
jsitarek, line 122: why did you change to str? this object is accessed as camera geometry object

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -227,21 +233,35 @@ def process_telescope_data(input_file, config, tel_id, camgeom, focal_eff):
...
+            camgeom[cleanmask],
+            corr_image[cleanmask],
+            peak_time[cleanmask],
+            hillas_params,
+        )
         leakage_params = leakage_parameters(camgeom, corr_image, cleanmask)
-        conc_params = concentration_parameters(camgeom, corr_image, hillas_params)
+        if assigned_tel_ids["LST-1"] == tel_id:
+            conc_params = concentration_parameters(
+                camgeom[cleanmask], image[cleanmask], hillas_params
```
jsitarek, line 249: it should be corr_image and not image

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -227,21 +233,35 @@ def process_telescope_data(input_file, config, tel_id, camgeom, focal_eff):
...
+            )
+        else:
+            conc_params = concentration_parameters(camgeom, image, hillas_params)
```
jsitarek, line 252: image ==> corr_image

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -227,21 +233,35 @@ def process_telescope_data(input_file, config, tel_id, camgeom, focal_eff):
...
-        timing_params = timing_parameters(camgeom, corr_image, peak_time, hillas_params)
+        timing_params = timing_parameters(
+            camgeom[cleanmask],
+            corr_image[cleanmask],
+            peak_time[cleanmask],
+            hillas_params,
+        )
         leakage_params = leakage_parameters(camgeom, corr_image, cleanmask)
-        conc_params = concentration_parameters(camgeom, corr_image, hillas_params)
+        if assigned_tel_ids["LST-1"] == tel_id:
```
jsitarek, line 247: please add a small comment before this line: "For LST-1 we compute concentration fro the cleaned image and for MAGIC from the full image to reproduce the current behaviour in the standard code.

jsitarek (approved):

PR #244, opened 2024-08-08
skip broken subruns

Elisa-Visentin:


Elisa-Visentin:
Currently NSB cannot be evaluated for 130 runs. I am trying to include some exceptions (subruns to be skipped) so that, hopefully, less jobs will crash and we could discard less runs 🤞 🤞

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/LSTnsb.py
```diff
@@ -65,12 +92,26 @@ def nsb(run_list, simtel, lst_config, run_number, denominator):
         if ii % mod == 0:
             try:
                 a, _, _ = calculate_noise_parameters(simtel, run_list[ii], lst_config)
-                noise.append(a)
-                logger.info(a)
+                if a is not None:
+                    if a > 0.0:
```
jsitarek, line 96: I do not think a negative output of calculate_noise_parameters is possible, see here:

https://github.com/cta-observatory/cta-lstchain/blob/378a4c3a8a1e72d8e3e1b6739edef9e591002261/lstchain/image/modifier.py#L393-L409

you get positive value if there is noise to be added, and if noise in data looks smaller than in MC, it is clipped to 0, but this is not a problem, just a very "dark" run. 
have you seen any cases with negative values?

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/LSTnsb.py
```diff
@@ -65,12 +92,26 @@ def nsb(run_list, simtel, lst_config, run_number, denominator):
...
+                if a is not None:
+                    if a > 0.0:
+                        noise.append(a)
+                        logger.info(a)
+                    else:
+                        mod = update_mod(
+                            mod, len(run_list), denominator, ii, len(noise)
+                        )
+                        logger.warning(
+                            f"NSB level could not be adequately evaluated for subrun {subrun} (negative value or missing pedestal events): skipping this subrun..."
```
jsitarek, line 104: negative value and missing pedestals are two completely different problems

jsitarek (commented):
it looks strange to me that you can get a negative output from calculation of the added NSB noise

Elisa-Visentin:
Just added a condition in the if not to run into a 'division by 0' error

Elisa-Visentin:
While checking all the logs looking for other minor issues.... These last commits only allow us to evaluate NSB for about 1/3 of the failed runs, w/o apparently changing the NSB value estimated for the other runs

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/LSTnsb.py
jsitarek: thx for checking it as we discussed, resolving comment

jsitarek (approved):

jsitarek:
I am fine with the current version, do you want to merge it now, or you want to commit anything else to the branch (since you mentioned that still 2/3 of the failed runs remain)

Elisa-Visentin:
@jsitarek Now approx 80 runs fail, but they have no FF events (not even for a single subrun), so there is no way to evaluate NSB for them (at least with the lstchain function, but since they are 2-3% of the total runs I think that we can discard them without trying to implement new algorithms). So I think that it can be merged (I also checked all the logs for the failed jobs and there seems to be no other minor 'bug fix' to be implemented)

PR #246, opened 2024-08-09
fix db paths

Elisa-Visentin:


jsitarek (approved):

jsitarek:
I approved it as it is, but it still reads two MAGIC+LST1 databases, so we might add some explanation why there are two files needed

PR #243, opened 2024-08-08
Auto mcp w/o MC

Elisa-Visentin:

PR #248, opened 2024-08-09
use tel id from config.yaml

Elisa-Visentin:


magicctapipe/scripts/lst1_magic/semi_automatic_scripts/coincident_events.py
```diff
@@ -62,7 +62,7 @@ def configfile_coincidence(target_dir, source_name, config_gen):
         config_dict = yaml.safe_load(fc)
 
     conf = {
```
jsitarek, line 64: with both the mc_tel_ids and event_coincidence taken from base_config_file you are basically using directly this base_config_file. So in principle we could dump the whole "config_file" into file_name (or even copy a file) to simplify the code, but having the config file stripped of all unnecessary settings also has its advantages, so perfectly fine for me to keep it like this as well.

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/dl1_production.py
```diff
@@ -76,7 +76,7 @@ def config_file_gen(target_dir, noise_value, NSB_match, source_name, config_gen)
         LST_config["increase_nsb"]["extra_bias_in_dim_pixels"] = noise_value[2]
         LST_config["increase_nsb"]["extra_noise_in_bright_pixels"] = noise_value[1]
     conf = {
```
jsitarek, line 78: similar like for the coincidence, we could save directly the config read from config_file.
for MCs you would modify it in the meantime in the lines 74-77, but those lines will be either way gone with MC part removed

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/stereo_events.py
```diff
@@ -64,7 +63,7 @@ def configfile_stereo(target_dir, source_name, config_gen):
     ) as fc:  # "rb" mode opens the file in binary format for reading
         config_dict = yaml.safe_load(fc)
     conf = {
-        "mc_tel_ids": config_gen["mc_tel_ids"],
+        "mc_tel_ids": config_dict["mc_tel_ids"],
         "stereo_reco": config_dict["stereo_reco"],
```
jsitarek, line 67: and one more in case you want to simplify the generation of configs

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/coincident_events.py
Elisa-Visentin: Indeed the idea was to split the (quite huge) config.yaml file in a set of subfiles (one per step), each one containing only the information needed in this step (both user-friendly and, since we have to open this configuration file in all the jobs to read it, maybe also a bit more 'SLURM-friendly')

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/dl1_production.py
Elisa-Visentin: I removed the change in the noise values. Then #243 should completely remove the MC

jsitarek: ok, but if you removed adding noise already in this PR, NSB_match does not make sense, and the whole MC part is missing will be missing one element after another. 
Actually I was not suggesting to remove the added noise already in this PR, sorry if it was not clear from my previous comment. Maybe best that you add this back (sorry again) and then we merge it.

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/coincident_events.py
jsitarek: ok, perfectly fine with me

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/dl1_production.py
Elisa-Visentin: No worries, my fault. I will add it again and check if the 'no MC' PR needs a similar fix in the config

jsitarek:
I think you also need to update the readme file:
https://github.com/cta-observatory/magic-cta-pipe/blob/40d483d637f1a363e55472dedfe55a470afd520d/magicctapipe/scripts/lst1_magic/README.md?plain=1#L36

Elisa-Visentin:
I updated the readme in the 'no_MC' PR (since some of the lines to be updated are the same, this way there should be no conflicts)

PR #251, opened 2024-09-02
Update requirements

aleberti:
Update requirements for conda package creation.


aleberti:
Merging now, tests now are passing, it took a long time because download at some point gets disrupted...

PR #252, opened 2024-09-06
In_out_versions

Elisa-Visentin:
Add the possibility to use, for each step, input data processed with an older MCP version wrt the current one

### DIFF
--- a/magicctapipe/scripts/lst1_magic/semi_automatic_scripts/coincident_events.py
+++ b/magicctapipe/scripts/lst1_magic/semi_automatic_scripts/coincident_events.py
@@ -66,7 +66,10 @@ def configfile_coincidence(target_dir, source_name, config_file):
         "event_coincidence": config_dict["event_coincidence"],
     }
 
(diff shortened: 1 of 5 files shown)


jsitarek:
I checked the files and the changes look ok. Once a test is run we should also check the job_accounting script if it is not missing any input files because they are in a different folder

Elisa-Visentin:
I am still testing it (IT queues are a bit stuck) but I will check both output and, as you suggested, job_accounting. Hopefully during the we...

Elisa-Visentin:
Actually there was a small issue with directories (I have to create the directory with the current version name to be able to store the config. file for stereo and coincidence, or the scripts cannot read it). Now it should be fixed

Elisa-Visentin:
As for job accounting, I plan to fix and check it today :)

jsitarek (approved):

jsitarek:
looks good, shall we merge?

Elisa-Visentin:
For me, yes (I finished all the tests and job accounting is fine)

PR #253, opened 2024-09-06
Ra-Dec

Elisa-Visentin:
Add ra-dec and MC dec information in LST database

### DIFF
--- a/magicctapipe/scripts/lst1_magic/README.md
+++ b/magicctapipe/scripts/lst1_magic/README.md
@@ -154,11 +154,11 @@ Eventually, to merge DL1 stereo (LST) subruns into runs, we run the `merge_stere
 
 > $ merge_stereo (-c config_auto_MCP.yaml)
 
-### Random forest and DL1 to DL2
+### DL1 to DL2
 
 TBD.
 
-### Instrument response function and DL3
+### DL2 to DL3
 
 TBD.
 


Elisa-Visentin:
This new scripts seems to work fine. We still miss some entries into the dictionaries, but we just have to update the txt files later on (while working on Dl2 and DL3)

aleberti:
Why do you need two files to specify the RA, Dec and point-like/extended property of the same sources? Can't you have e.g.  a single yaml file to store all that info? Or also JSON, since you use it anyway to open the txt files.

Elisa-Visentin:
> Why do you need three files to specify the RA, Dec and point-like/extended property of the same sources? Can't you have e.g. a single yaml file to store all that info? Or also JSON, since you use it anyway to open the txt files.

A single file with 3 dictionaries inside? Maybe quite a mess when we have to modify it (hundreds of lines and quite difficult to find out if a source is already there in which dictionary) 
We could store each entry as a list ('source':[ra, dec, point]) but then some of the entries of the list will be null (not known or not needed): what about this solution?

Elisa-Visentin:
We also have a problem with Crab (during one of the obs. they used 'Crab' instead of 'CrabNebula' and astropy returns two different coordinates, while I think that they should have the same coords.
image

aleberti:
Yes, I meant the solution you proposed i.e. one list for each source, and each source name becomes a key.

Regarding the Crab/CrabNebula, it is likely our problem to solve. 99% I am sure that the configuration on the LST side have the same coordonates. Astropy gives slightly different ones because most probably Crab is the pulsar (not sure which catalog is used though).

Elisa-Visentin:
I'm almost sure that the source is always the same (as for the observations): probably they just put the wrong name. astropy retrieves coordinates from simbad usually. This can be easily solved in this script by adding an 'if', but I will also check MAGIC root files for one of these obs (just to be sure that the coordinates are the same)

aleberti:
> We also have a problem with Crab (during one of the obs. they used 'Crab' instead of 'CrabNebula' and astropy returns two different coordinates, while I think that they should have the same coords.

here are we talking about MAGIC or LST? in LST I see only one configuration called `crab` (plus other configurations with different wobble offset, but with very explicative names). Of course I cannot exclude that there was another configuration with a different name that was deleted and therefore not listed anymore.

Elisa-Visentin:
Now I am checking MAGIC coordinates (easier: they are written in the files) and there seem to exist two different sources (Crab and CrabNebula) according to MARS and data taking: also in the runbook there is a different name in schedule wrt run summary

aleberti:
@Elisa-Visentin can you tell one of these dates? I would like to check better, because in the scheduling catalog files I did not find source with the name "Crab", only "CrabNebula".

Elisa-Visentin:
OK, from MAGIC:
CrabNebula: 22° 00′ 52″,   05:34:32 (h:m:s)   deg:    22.0144444    83.6333333          

Crab: 22° 00′ 51",  05:34:31 (h:m:s)     deg:  22.0141667     83.6291667    

mhhh... the second one does not even correspond to the one by astropy

Elisa-Visentin:
3-4 March 2024 (LST): Yes the name is the 'wrong' one in run summary, not in scheduling

Elisa-Visentin:
In the mean time, I fixed the input dictionary

aleberti:
ok, apparently there is a catalog of sources in MAGIC called "favorites" where indeed we have:

```bash
Crab,f|T,+5:34:31.00,+22:0:51.00,1.0,2000
```

that's what shifters used.

aleberti:
> * we ignore the sec. difference and put the same coordinates (which?) for both of them to be able to stack data?

effectively for observations that difference does not affect analysis. Please use the coordinates of `CrabNebula`

> * we keep the two sources as different and fix the Crab coordinates to the ones actually used during the obs.?

I already modified the *only* catalog file of MAGIC where the name was `Crab` instead of `CrabNebula`. So now it is:

```bash
CrabNebula,f|T,5:34:32.00,22:00:52.00,1.0,2000
```

so, it was just really a remote coincidence that shifters used that catalog...

Elisa-Visentin:
fixed. Thanks

aleberti (approved):

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/set_ra_dec.py
```diff
@@ -0,0 +1,171 @@
+"""
+Add, to LST database, infos about coordinates and extension of the sources and MC declination to be used to process the source
+
+Usage:
+$ set_ra_dec (-b YYYYMMDD -e YYYYMMDD -r ra_dict -d dec_dict -m mc_dec -p point_source_dict)
```
jsitarek, line 5: options -r -d and -m were changes into one -s

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/set_ra_dec.py
```diff
@@ -0,0 +1,171 @@
...
+    i = 0
+    for src in sources:
+
+        try:
+            coord = SkyCoord.from_name(src)
+            if src == "Crab":
+                coord = SkyCoord.from_name("CrabNebula")
+            src_dec = coord.dec.degree
+            src_ra = coord.ra.degree
+            df_LST["ra"] = np.where(df_LST["source"] == src, src_ra, df_LST["ra"])
```
jsitarek, line 108: those 3 lines are also repeated in the except block of the code. I think you can put them once after the except block always setting them to src_ra, src_dec and min dec if those are available, and if not to Nan

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/set_ra_dec.py
```diff
@@ -0,0 +1,171 @@
...
+            df_LST["dec"] = np.where(df_LST["source"] == src, src_dec, df_LST["dec"])
+            df_LST["MC_dec"] = np.where(
+                df_LST["source"] == src,
+                float(dec_mc[np.argmin(np.abs(src_dec - dec_mc))]),
+                df_LST["MC_dec"],
+            )
+        except NameResolveError:
+            print(f"{i}: {src} not found in astropy. Looking to the dictionaries...")
+            if (src in source_dict.keys()) and (source_dict.get(src)[0] != "NaN"):
```
jsitarek, line 117: maybe you can read RA and Dec together? they should be both either available or both NaN, and it would simplify the code

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/source_dict.json
```diff
@@ -0,0 +1,125 @@
...
+"MAXIJ1848-01":[282.2, -1.4867, "NaN"],
+"1ES1028+51":[157.82718791991, 50.89328257724, "True"],
+"1A-0535+262":[84.72739538276, 26.31578804244, "NaN"],
+"Cyg-X3":[308.107417, 40.95, "True"],
+"S20109+22":[18.0241667, 22.73333, "True"],
+"S30218+35":[35.27279, 35.93333, "True"],
+"LHAASOJ2108+":[317.15000000000003, 51.95, "False"],
+"LhaasoJ1956+":[299.05, 28.75, "False"],
+"PSRJ2229+611":[337.271925, 61.2356889, "True"],
+"1LHAASOJ0007":[1.7541666666666667, 73.05194444444444, "True"],
```
jsitarek, line 38: this is probably an extended source

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/source_dict.json
```diff
@@ -0,0 +1,125 @@
...
+"IC-220303A":[267.79999999999995, 11.42, "NaN"],
+"RGB0521+211":[80.44152391208, 21.21429244076, "True"],
+"FermiJ0330.1":[52.50833333333333, 37.73, "NaN"],
+"HESSJ1912":[288.26666666666665, 10.207777777777778, "False"],
+"Abell-1132-M":[164.68333333333334, 56.592777777777776, "NaN"],
+"Cygnus-XNort":[310.0416666666667, 42.471944444444446, "NaN"],
+"GBR230512A":[295.39583333333337, 38.39972222222222, "True"],
+"GRB220603":[284.15, 18.84, "True"],
+"HESSJ1848-01":[282.12083333333334, -1.0527777777777778, "False"],
+"Perseus-MA":[49.5625, 41.418055555555554, "NaN"],
```
jsitarek, line 59: this is not a real source but a position in between NGC 1275 and IC 310, so you can put "False" to point like

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/source_dict.json
```diff
@@ -0,0 +1,125 @@
+{"TCrB":[239.8756667, 25.91667, "NaN"],
```
jsitarek, line 1: it would be useful to sort this list somehow: by source name, by RA, or having sources with RA/Dec specified before the ones with NaNs. Now it looks like the sequence is rather random

jsitarek (changes requested):
I put a few small comments

PR #259, opened 2024-09-12
ra dec db

Elisa-Visentin:
Now it is 'safe'. I will list comments/requests from the old PR below

### DIFF
--- a/magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/set_ra_dec.py
+++ b/magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/set_ra_dec.py
@@ -105,40 +105,33 @@ def main():
                 coord = SkyCoord.from_name("CrabNebula")
             src_dec = coord.dec.degree
             src_ra = coord.ra.degree
-            df_LST["ra"] = np.where(df_LST["source"] == src, src_ra, df_LST["ra"])
-            df_LST["dec"] = np.where(df_LST["source"] == src, src_dec, df_LST["dec"])
+
+        except NameResolveError:
+            print(f"{i}: {src} not found in astropy. Looking to the dictionaries...")
+            if (
+                (src in source_dict.keys())
+                and (source_dict.get(src)[0] != "NaN")
+                and (source_dict.get(src)[1] != "NaN")
+            ):
+                src_ra = float(source_dict.get(src)[0])
+                src_dec = float(source_dict.get(src)[1])
+
+            else:
+                print(
+                    f"\t {i}: {src} RA and/or Dec not in the dictionary. Please update the dictionary"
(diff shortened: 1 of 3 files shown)


Elisa-Visentin:
- [ ] options changed to -s (doc string)
- [ ] lines to set ra, dec, mc dec repeated in try except
- [ ] read ra and dec together from dictionary
- [ ] one extended source in db and one 'not source'
- [ ] sort dictionary

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/set_ra_dec.py
```diff
@@ -0,0 +1,171 @@
...
+    i = 0
+    for src in sources:
+
+        try:
+            coord = SkyCoord.from_name(src)
+            if src == "Crab":
+                coord = SkyCoord.from_name("CrabNebula")
+            src_dec = coord.dec.degree
+            src_ra = coord.ra.degree
+            df_LST["ra"] = np.where(df_LST["source"] == src, src_ra, df_LST["ra"])
```
jsitarek, line 108: adding again the comment from the original PR:
those 3 lines are also repeated in the except block of the code. I think you can put them once after the except block always setting them to src_ra, src_dec and min dec if those are available, and if not to Nan

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/set_ra_dec.py
```diff
@@ -0,0 +1,171 @@
...
+            df_LST["dec"] = np.where(df_LST["source"] == src, src_dec, df_LST["dec"])
+            df_LST["MC_dec"] = np.where(
+                df_LST["source"] == src,
+                float(dec_mc[np.argmin(np.abs(src_dec - dec_mc))]),
+                df_LST["MC_dec"],
+            )
+        except NameResolveError:
+            print(f"{i}: {src} not found in astropy. Looking to the dictionaries...")
+            if (src in source_dict.keys()) and (source_dict.get(src)[0] != "NaN"):
```
jsitarek, line 117: adding again the comment from the previous PR:

maybe you can read RA and Dec together? they should be both either available or both NaN, and it would simplify the code

Elisa-Visentin: should be better now

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/set_ra_dec.py
Elisa-Visentin: should be fixed now

PR #260, opened 2024-09-13
Adding a function to create a log file and save cloud correction para…

nzywucka:
…meters in the file

### DIFF
--- a/magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
+++ b/magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
@@ -12,8 +12,10 @@ $ python lst_m1_m2_cloud_correction.py
 """
 import argparse
 import logging
+import re
 
 import astropy.units as u
+import ctapipe
 import numpy as np
 import pandas as pd
 import yaml
@@ -28,6 +30,7 @@ from ctapipe.image import (
 from ctapipe.instrument import SubarrayDescription
 from ctapipe.io import read_table
 
+import magicctapipe
 from magicctapipe.io import save_pandas_data_in_table
 
 logger = logging.getLogger(__name__)
@@ -35,9 +38,41 @@ logger.addHandler(logging.StreamHandler())
 logger.setLevel(logging.INFO)
 
 
+def save_cloud_correction_params(config, run_number):
+    """
+    Creates a log file and saves the cloud correction parameters from the config file and software information in the log file
+
+    Parameters
+    ----------
+    config : dict
+        Configuration for the LST-1 + MAGIC analysis
+    run_number : str
+        Number of the analysed run
+    """
+    cloud_correction = config.get("cloud_correction", {})
+    log_filename = f"correction_log_{run_number}.log"
+    logging.basicConfig(
+        filename=log_filename,
+        level=logging.INFO,
+        format="%(asctime)s - %(levelname)s - %(message)s",
+    )
+
+    logging.info(f"Cloud Correction Parameters for Run {run_number}:")
+    logging.info(f"{cloud_correction}")
+
+    with open(f"cloud_correction_params_{run_number}.yaml", "w") as f:
+        yaml.dump(cloud_correction, f, default_flow_style=False)
+
+    ctapipe_version = ctapipe.__version__
+    magicctapipe_version = magicctapipe.__version__
+
+    logging.info(f"ctapipe version: {ctapipe_version}")
+    logging.info(f"magicctapipe version: {magicctapipe_version}")
+
+
 def model0(imp, h, zd):
     """
-    Calculated the geometrical part of the model relating the emission height with the angular distance from the arrival direction
+    Calculates the geometrical part of the model relating the emission height with the angular distance from the arrival direction
 
     Parameters
     ----------
@@ -340,8 +375,13 @@ def main():
         else:
             raise ValueError(f"No optics data found for telescope: {telescope.name}")
 
+    run_number_match = re.search(r"Run\d+\.\d+", args.input_file)
+    run_number = run_number_match.group() if run_number_match else "unknown_run"
+
     with open(args.config_file, "r") as file:
         config = yaml.safe_load(file)
+
+    save_cloud_correction_params(config, run_number)
 
     tel_ids = config["mc_tel_ids"]
 


jsitarek:
I think you did it well too complicated. You already have one instance of "logging": logger that can be used, you just need to print out in the general log cloud_correction parameters from the config file (and also you can print out MCP  version if you want, but this I think we should change such that it is done for all the scripts)

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -248,7 +252,7 @@ def process_telescope_data(input_file, config, tel_id, camgeom, focal_eff):
         if assigned_tel_ids["LST-1"] == tel_id:
             conc_params = concentration_parameters(
                 clean_camgeom, clean_image, hillas_params
-            )  # "For LST-1 we compute concentration from the cleaned image and for MAGIC from the full image to reproduce the current behaviour in the standard code
+            )  # "For LST-1 we compute concentration from the cleaned image and for MAGIC from the full image to reproduce the current behaviour in the standard code/home/natti/CTA/lodz_cluster/lstchain_10_11_test/scripts/new/outfiles/dl1_coincidence_stereo_0.75-0.85_03955
```
jsitarek, line 255: this looks like a copy/paste error

magicctapipe/scripts/lst1_magic/lst_m1_m2_cloud_correction.py
```diff
@@ -379,11 +384,32 @@ def main():
...
+    logging.basicConfig(
+        filename=log_filepath,
+        level=logging.INFO,
+        format="%(asctime)s - %(levelname)s - %(message)s",
     )
 
-    subarray_info.to_hdf(args.output_file)
+    logging.info(f"Correction parameters: {correction_params}")
+    logging.info(f"ctapipe version: {ctapipe.__version__}")
+    logging.info(f"magicctapipe version: {magicctapipe.__version__}")
```
jsitarek, line 412: why do you use a separate log? you have "logger" defined before and used just below this line. you can use it also to write the correction parameters and version of software

jsitarek:
Hi @nzywucka, the latest changes seem to went into another file.

jsitarek:
Hi @nzywucka 
With the fix of Alessio I rerun the tests and now they are successful, but there are still multiple unresolved comments (and somewhat on the way you changed the name of the script, so you have some of the changes implemented in a second file).

magicctapipe/resources/config.yaml
```diff
@@ -102,7 +102,7 @@ event_coincidence:
 
 
 stereo_reco:
-    quality_cuts: "(intensity > 50) & (width > 0)"
+    quality_cuts: "(intensity > 50) & (intensity > 80 or tel_id > 1) & (width > 0)" #"(intensity > 50) & (width > 0)"
```
jsitarek, line 105: this cut was used for tests. it should not go into the default input card

magicctapipe/resources/config.yaml
```diff
@@ -280,8 +280,8 @@ dl2_to_dl3:
     
 cloud_correction:
     specify_cloud: yes
```
jsitarek, line 282: it is not anymore "specify_cloud" option and it is not used anywhere in the code. I think you can drop it for the moment

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,425 @@
+#!/usr/bin/env python
```
jsitarek, line 1: there is still a mess up with some of the correction in the previous file and added a second file. this needs to be cleared up before the code can be merged.

magicctapipe/resources/config.yaml
```diff
@@ -279,9 +279,22 @@ dl2_to_dl3:
...
-    specify_cloud: yes
-    base_height: "5900. m" 
-    thickness: "1320. m"
-    vertical_transmission: 1.0
+    max_gap_lidar_shots: "900 s"
+    lidar_report_file: "../../resources/lidar_reports_clouds_ST.03.16.yaml"
     number_of_layers: 10
 
+    additional_lst_cleaning:
+        picture_thresh: 10
```
jsitarek, line 287: the purpose of redoing the cleaning is to raise the thresholds by some factor, and the same factor should be applied to LST and MAGIC. So, either the factor is determined automatically in the script (but this can be messy because then one needs to process the simulations for each run even if changes are small, or you specify in the input card only a single multiplication factor and this factor is applied to picture_thresh, boundary_thresh of both MAGIC and LST, and rest of the settings should be taken from default config or from the file itself (at least for MAGIC, because for LST this can be more tricky)

magicctapipe/resources/lidar_reports_clouds_ST.03.16.yaml
```diff
@@ -0,0 +1,533 @@
+units:
+  timestamp: 'iso'
```
jsitarek, line 2: this file is for a very specific dataset, therefore there is no point of posting it into the general repository

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,715 @@
...
+        - "base_height", "top_height", "transmission", "lidar_zenith"
+
+    Returns
+    --------
+    tuple or None
+
+        A tuple containing interpolated or nearest values for:
+        - base_height (float): The base height of the cloud layer in meters.
+        - top_height (float): The top height of the cloud layer in meters.
+        - vertical_transmission (float): Transmission factor of the cloud layer adjusted for vertical angle.
```
jsitarek, line 147: ... for zenith angle

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,715 @@
...
+        Data frame of corrected DL1 parameters
+    """
+
+    # AC LST
+    cleaning_level = config["LST"]["tailcuts_clean"]
+    # Multiply only non-boolean numerical values
+    config_clean_LST = {
+        key: (
+            value * cmf
+            if isinstance(value, (int, float)) and not isinstance(value, bool)
```
jsitarek, line 308: this way you will upscale all the parameters, including e.g. min_number_picture_neighbours that should not be touched. It is better and simpler to directly upscale only picture_thresh and boundary_thresh

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,715 @@
...
+            else value
+        )
+        for key, value in cleaning_level.items()
+    }
+
+    # AC MAGIC
+    # Ignore runtime warnings appeared during the image cleaning
+    warnings.simplefilter("ignore", category=RuntimeWarning)
```
jsitarek, line 316: even while this is copied from the magic_calib_do_dl1 I think it would be better to try to run without this and see if there are no issues. Because it might also hide some other warnings related to different parts of the code

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,715 @@
...
+    # Configure the MAGIC image cleaning
+    config_clean = config["MAGIC"]["magic_clean"]
+    # Scale numeric values and set 'find_hotpixels' to False
+    config_clean_magic = {
+        key: (
+            False
+            if key == "find_hotpixels"
```
jsitarek, line 323: save as above, you should only scale picture_thresh and boundary_thresh, and do not touch other parameters

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,715 @@
...
+
+        if len(inds_img) == 0:
+            raise ValueError("Error: 'inds_img' list is empty!")
+        index_img = inds_img[0]
+        image = dl1_images["image"][index_img]
+        clean_mask = dl1_images["image_mask"][index_img]
+        peak_time = dl1_images["peak_time"][index_img]
+        image /= trans_pixels
+
+        clean_image = image[clean_mask]
```
jsitarek, line 428: you use the old mask first no matter if you do the additional cleaning a moment later or not. I think it is faster to move those 3 lines as "else" after the next condition

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,715 @@
...
+            else:
+
+                # Apply the image cleaning
+                clean_mask, image, peak_time = magic_clean.clean_image(
+                    event_image=image,
+                    event_pulse_time=peak_time,
+                    unsuitable_mask=unsuitable_mask,
+                )
+
+                # n_pixels = np.count_nonzero(clean_mask)
```
jsitarek, line 466: some leftover commented out lines here

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,715 @@
...
+        event_params = {
+            **hillas_params,
+            **timing_params,
+            **leakage_params,
+        }
+
+        prefixed_conc_params = {
+            f"concentration_{key}": value for key, value in conc_params.items()
+        }
+        event_params.update(prefixed_conc_params)
```
jsitarek, line 498: prefixed_conc_params could not be included directly in event_params a few lines above?

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,715 @@
...
+        optics_row = optics_table[optics_table["optics_name"] == telescope.name]
+        if len(optics_row) > 0:
+            focal_eff[telid] = optics_row["effective_focal_length"][0] * u.m
+        else:
+            raise ValueError(f"No optics data found for telescope: {telescope.name}")
+
+    with open(args.config_file, "r") as file:
+        config = yaml.safe_load(file)
+
+    tel_ids = config["mc_tel_ids"]
```
jsitarek, line 581: why do you get them from here? those are MC tel_ids, you should be able to use the tel_descriptions above to figure out what telescope you are looking at now

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,715 @@
...
+    cmf = correction_params.get("cleaning_multiplication_factor")
+
+    dl1_params = read_table(args.input_file, "/events/parameters")
+
+    mean_subrun_timestamp = np.mean(dl1_params["timestamp"])
+    mean_subrun_zenith = np.mean(90.0 - np.rad2deg(dl1_params["pointing_alt"]))
+
+    cloud_params = lidar_cloud_interpolation(
+        mean_subrun_timestamp, max_gap_lidar_shots, lidar_report_file
+    )
```
jsitarek, line 600: this function can return None if there are no proper LIDAR raports to interpolate from that would crash the code below. A better way to handle error management would be to introduce a new error here:
https://github.com/cta-observatory/magic-cta-pipe/blob/73e4a3bf47ace47deb11d1d74a4caf0c37954841/magicctapipe/utils/error_codes.py
and throw it like here: https://github.com/cta-observatory/magic-cta-pipe/blob/73e4a3bf47ace47deb11d1d74a4caf0c37954841/magicctapipe/scripts/lst1_magic/lst1_magic_stereo_reco.py#L234

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,715 @@
...
+
+    with open(args.config_file, "r") as file:
+        config = yaml.safe_load(file)
+
+    tel_ids = config["mc_tel_ids"]
+
+    correction_params = config.get("cloud_correction", {})
+    max_gap_lidar_shots = u.Quantity(
+        correction_params.get("max_gap_lidar_shots")
+    )  # set it to 900 seconds
```
jsitarek, line 586: the comment is misleading - the code will set it to whatever is in the input card

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,715 @@
...
+    dfs = []
+
+    for tel_name, tel_id in tel_ids.items():
+        if tel_id != 0:  # Only process telescopes that have a non-zero ID
+            # Read images for each telescope
+            image_node_path = "/events/dl1/image_" + str(tel_id)
+            try:
+                dl1_images = read_table(args.input_file, image_node_path)
+            except tables.NoSuchNodeError:
+                logger.info(f"\nNo image for telescope with ID {tel_id}. Skipping.")
```
jsitarek, line 616: this skips the telescopes without stored images and it is very dangerous because images are stored separately in MAGIC and in LST files, so e.g. if you have LST file without images, but MAGIC with images the script will make MAGIC-only analysis that is likely not what the user wanted. 
So I would turn this into a fatal error

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,715 @@
...
+                dl1_images = None
+
+            df = process_telescope_data(
```
jsitarek, line 619: you pass the config to the function, so the parameters like tel_ids, nlayers, cmf should be recovered from the config inside this function instead of being passed additionally.

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,715 @@
...
+        cloud_vertical_trans_array = f.create_array(
+            cloud_metadata_group,
+            "cloud_vertical_transmission",
+            np.array([cloud_vertical_trans_data]),
+        )
+        cloud_vertical_trans_array.attrs["unit"] = str(u.dimensionless_unscaled)
+
+    subarray_info.to_hdf(output_file)
+
+    correction_params = config.get("cloud_correction", {})
```
jsitarek, line 701: you already extracted it in L583, why do you do it again?

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
```diff
@@ -0,0 +1,715 @@
...
+
+    dl1_params = read_table(args.input_file, "/events/parameters")
+
+    mean_subrun_timestamp = np.mean(dl1_params["timestamp"])
+    mean_subrun_zenith = np.mean(90.0 - np.rad2deg(dl1_params["pointing_alt"]))
+
+    cloud_params = lidar_cloud_interpolation(
+        mean_subrun_timestamp, max_gap_lidar_shots, lidar_report_file
+    )
+
```
jsitarek, line 601: also I think it would be useful to print the cloud_params in the logger

magicctapipe/scripts/lst1_magic/extracted_lidar_reports_test.csv
```diff
@@ -0,0 +1,7 @@
+timestamp,base_height,top_height,transmission,lidar_zenith
```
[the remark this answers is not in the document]
jsitarek: this comment is still valid

magicctapipe/resources/lidar_reports_clouds_ST.03.16.yaml
jsitarek: this comment is still valid

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
jsitarek: still valid

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
jsitarek: still valid

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
jsitarek: still valid

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
jsitarek: still valid question

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
jsitarek: still valid

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
jsitarek: still valid

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
jsitarek: still valid

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
jsitarek: still very much valid

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_correction.py
jsitarek: still valid

PR #263, opened 2024-09-23
fix combo in DL2 real data nb

Elisa-Visentin:


Elisa-Visentin:
Since this is only a minor fix and Seiya already approved it, I will merge it now and also close the related issue

PR #265, opened 2024-10-04
Pin version of micromamba.

aleberti:
Tests are failing because of a problem in micromamba, see https://github.com/mamba-org/setup-micromamba/issues/225. Version is now pinned, to be removed when the problem is solved on micromamba's side.


aleberti:
Closing this as it is solved in #266 by using the latest version of micromamba action released some hours ago.

PR #270, opened 2024-11-20
Mc dense

Elisa-Visentin:
Add the possibility to use MC line with high density for training for some specific sources (till now I am sure only about Boomerang, whose results have already been published; if other sources still not public must be added, maybe we can retrieve their names from a private file)


jsitarek:
I think for Boomerang they are either way going to do something special with a mixture of standard and dense line

Elisa-Visentin:
I contacted FC: trey do the training with dense, then, for diffuse analysis, they use the non-dense for full-enclosure IRFs, so they could (try to) at least use our DL2s if we produce them with dense (as for Dl3, they will have only the point-like ones from us, but they can produce their own with their IRFs)

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/DL1_to_DL2.py
```diff
@@ -4,7 +4,7 @@
 the data level 2.
 
 Usage:
-$ DL1_to_DL2 -c configuration_file.yaml
+$ DL1_to_DL2 -c configuration_file.yaml -d list_dense.txt
```
jsitarek, line 7: this -d option is optional as far as I can see (and should be) so you can put it here as [-d list_dense.txt]

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/DL2_to_DL3.py
```diff
@@ -178,7 +184,20 @@ def main():
...
+        dest="dense_list",
+        type=str,
+        help="File with name of sources to be processed with the dense MC train line",
+    )
+
     args = parser.parse_args()
+    dense_list = []
+    if args.dense_list is not None:
+        with open(args.dense_list) as d:
+            dense_list = d.readlines()
```
jsitarek, line 199: if you do like this, the list will have '\n' at the end of each position, and then the source names will not be matched (unless they include '\n' as well) - have you checked if this works?
you can get read of '\n' easily if you do this instead:
```
dense_list=d.read().splitlines()
```

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/DL1_to_DL2.py
```diff
@@ -183,8 +190,22 @@ def main():
...
+        type=str,
+        help="File with name of sources to be processed with the dense MC train line",
+    )
+
 
     args = parser.parse_args()
+    dense_list = []
+    if args.dense_list is not None:
+        with open(args.dense_list) as d:
+            dense_list = d.readlines()
```
jsitarek, line 206: if you do like this, the list will have '\n' at the end of each position, and then the source names will not be matched (unless they include '\n' as well) - have you checked if this works?
you can get read of '\n' easily if you do this instead:
```
dense_list=d.read().splitlines()
```

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/DL2_to_DL3.py
```diff
@@ -4,7 +4,7 @@
 the data level 3.
 
 Usage:
-$ python new_DL2_to_DL3.py -c configuration_file.yaml
+$ python new_DL2_to_DL3.py -c configuration_file.yaml -d list_dense.txt
```
jsitarek, line 7: [-d list_dense.txt]

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/DL1_to_DL2.py
```diff
@@ -137,6 +142,8 @@ def bash_DL1Stereo_to_DL2(target_dir, source, env_name, cluster, RF_dir, df_LST,
             dec = df_LST[df_LST.source == source].iloc[0]["MC_dec"]
             dec = str(dec).replace(".", "")
             RFdir = f"{RF_dir}/{period}/NSB{nsb}/{MC_v}/dec_{dec}/"
```
jsitarek, line 144: if you want you could make a shorthand if statement instead:
```
RFdir = f"{RF_dir}/{period}/NSB{nsb}/{MC_v}/dec_{dec}{'_high_density' if source in dense_list else ''}/"
````
it avoids code duplication and managing the same path in two places, but the line gets more complicated, so if you prefer the original it is also perfectly fine with me

File: magicctapipe/scripts/lst1_magic/semi_automatic_scripts/DL2_to_DL3.py [hunk 6]
```diff
@@ -129,6 +133,8 @@ def DL2_to_DL3(target_dir, source, env_name, IRF_dir, df_LST, cluster, MC_v):
             dec = df_LST[df_LST.source == source].iloc[0]["MC_dec"]
             dec = str(dec).replace(".", "")
             IRFdir = f"{IRF_dir}/{period}/NSB{nsb}/GammaTest/{MC_v}/g_dyn_0.9_th_glo_0.2/dec_{dec}/"
```
jsitarek, line 135: as above, if you want you can change to:
```
IRFdir = f"{IRF_dir}/{period}/NSB{nsb}/GammaTest/{MC_v}/g_dyn_0.9_th_glo_0.2/dec_{dec}{'_high_density' if source in dense_list else ''}/"
```

jsitarek (commented):
there is a merge conflict that needs to be fixed first, other than this it is fine, however I also left some suggestions

Elisa-Visentin:
I also added here some fixes to have more uniform date formats so that we can reduce a bit the parser lines

PR #273, opened 2024-11-27
Fixed DL3 creation complaining about mismatch in binning.

jsitarek:
It sometimes happened that the IRF files have different bin ranges saved despite using the same input card. The difference is very small, comparable to the numerical accuracy of double precision. I saw it in the files processed in CNAF and my suspicion is that it comes from inhomogenuity of machines over which the code was run. Now if different binning is detected, it checks also if the relative difference exceeds somewhere 10^-15 and only then raises an error.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -28,3 +28,6 @@ repos:
                 .*__init__.py|
                 .*/tests/.*
             )$
+
+default_language_version:
+    python: python3.11
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -30,4 +30,4 @@ repos:
             )$
 
 default_language_version:
-    python: python3.11
+    python: python3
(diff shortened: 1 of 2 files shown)


Elisa-Visentin:
Seems fine to me. Why the changes in pre-commit config? 
As for tests, I will work again on coverage and new tests after finishing auto-mcp (not to have too many open projects), so we can probably neglect it

jsitarek:
pre-commit was failing when I was trying to commit the change to github. For some reason it was trying to use python3.1, no idea why. This solved the problem (and allowed me to commit the changes) and I think it is harmless because we also require python 3.11 in the environment. But I am open to other solutions as well.

Elisa-Visentin:
I was just curious (there were no failings in the other PRs, and you just added an if/else, so this is quite strange). BTW, if it works, it is fine

Elisa-Visentin (approved):

PR #280, opened 2025-02-12
Implement custom energy-dependent cuts for IRFs and DL3 creations

gabemery:
These changes introduce the option to provide energy-dependent cuts on gammaness or theta  during the IRFs creation. And provides associated metadata in the IRFs and DL3 files.

Why is it needed : Allows to provide energy dependent cuts based on criteria different from gamma efficiency. Such as best sensitivity.

Usage : In the config file


Elisa-Visentin:
Thank you @gabemery. Very useful. I was just checking this quickly: there are if clauses inside an elif? (Maybe we can reformulate a bit this and through elifs). As for the new io function, I would suggest passing only the dictionary and string, instead of the whole config, so that we can try to keep backward compatibility (i.e., if the new 'entry' is not in the config, the script can exit w/o erros in the module), but this depends on if we want to create a new major version or not

gabemery:
> As for the new io function, I would suggest passing only the dictionary and string, instead of the whole config,

This is possible without issue, it is only moving config["interpolate_kind"], and config["custom_cuts"] as input to the function instead of config. 

> so that we can try to keep backward compatibility (i.e., if the new 'entry' is not in the config, the script can exit w/o erros in the module), but this depends on if we want to create a new major version or not

For backward compatibility I think it is already ok, since the new entries are only read in the new io function, and in log.info messages in the if blocks for "custom" cuts.
Also, if the method selected is custom, and the associated entries are missing, we should have an error. And looking into the config in the io function or before would lead to the same failure.

So I think there is nothing to change there.

Elisa-Visentin:
> Hi @Elisa-Visentin , Thank you for the review. I was also a bit bothered by the elif : if, but I didn't have a clear idea of a better solution.
> 
> Currently the structure is:
> 
> ```
> if global
>     apply cut, do metadata
> elif energy dependent
>     if dynamic
>         get dynamic cut and setup metadata
>     if custom
>         get custom cut and setup meta
>     apply energy dependent cut + other common code (35 lines)
> else
>     handle wrong keyword
> ```
> 
> To avoid duplicating 35 lines of code, for the energy depedent case, it is needed to have the cuts recovery handled for both before the application. To handle the wrong key word, everything needs to be handled in the same if, elif, else.
> 
> I can think of alternatives, but I am not sure which would be good. To give a few I can invent now :
> 
> 1. Handle the error first, do only if
> 
> ```
> if not global or dynamic or custom
>     error
> if global
>     apply cut, do metadata
> if dynamic
>      get dynamic cut and setup metadata
> if custom
>     get custom cut and setup meta
> if energy dependent (dynamic or custom)
>     apply energy dependent cut + other common code
> ```
> 
> 2. Duplicate the code or make a function handling it. Keep current if elif else
> 
> ```
> if global
>     apply cut, do metadata
> elif dynamic
>     get dynamic cut and setup metadata
>     apply energy dependent cut + other common code (35 lines)
> elif custom
>     get custom cut and setup meta
>     apply energy dependent cut + other common code (35 lines)
> else
>     handle wrong keyword
> ```

Hi, probably the most 'beautiful' solution is the last one, where a new function (which could also be defined inside the main function), so that we have only one if-elif-else statement, but I would suggest to wait for Alessio and Julian before changing this (maybe they have a better solution)

Elisa-Visentin:
> > As for the new io function, I would suggest passing only the dictionary and string, instead of the whole config,
> 
> This is possible without issue, it is only moving config["interpolate_kind"], and config["custom_cuts"] as input to the function instead of config.
> 
> > so that we can try to keep backward compatibility (i.e., if the new 'entry' is not in the config, the script can exit w/o erros in the module), but this depends on if we want to create a new major version or not
> 
> For backward compatibility I think it is already ok, since the new entries are only read in the new io function, and in log.info messages in the if blocks for "custom" cuts. Also, if the method selected is custom, and the associated entries are missing, we should have an error. And looking into the config in the io function or before would lead to the same failure.
> 
> So I think there is nothing to change there.

I have to check this better.... (sorry, I'm quite busy and I'm working on too many things at the same time). Probably we should first fix the nested ifs, to be sure about where we call the involved function: if it is only in the 'elif custom', it should work fine

magicctapipe/io/io.py
```diff
@@ -249,6 +251,53 @@ def format_object(input_object):
...
 
 
+def get_custom_cuts(config, energy_bins):
+    """
+    Extract and interpolate custom, energy-dependent cuts for the IRFs creation.
+
+    Parameters
+    ----------
+    config : dict
+        Dictionary containing the custom cuts and the interpolation tragedy.
```
jsitarek, line 261: while nearest point interpolation is not very precise I would not call it "tragedy" ;-), I guess you meant "strategy"

magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py
```diff
@@ -335,34 +346,47 @@ def create_irf(
             mask_gh = event_table_bkg["gammaness"] > cut_value_gh
             event_table_bkg = event_table_bkg[mask_gh]
 
-    elif cut_type_gh == "dynamic":
-        config_gh_cuts.pop("global_cut_value", None)
-
-        logger.info("\nDynamic gammaness cuts:")
-        logger.info(format_object(config_gh_cuts))
+    elif cut_type_gh == "dynamic" or "custom":
+        if cut_type_gh == "dynamic":
```
jsitarek, line 350: about the blocks pointed out by Elisa, 
you have now 
```
elif dynamic or custom
  if dynamic
    <dynamic code>
  if custom
    <custom code>
  <common code>
'''
you could lower the if level (at the same time lowering the number of modified lines) without making the code longer if you do like this:
'''
if dynamic
  <dynamic code>
elif custom
  <custom code>
if dynamic or custom
  <common code>
'''

File: magicctapipe/scripts/lst1_magic/lst1_magic_create_irf.py [hunk 3]
```diff
@@ -422,34 +446,56 @@ def create_irf(
...
-                values=event_table_gamma["theta"],
-                bin_values=event_table_gamma["reco_energy"],
-                bins=energy_bins,
-                fill_value=theta_cut_max,
-                percentile=theta_percentile,
-                min_value=theta_cut_min,
-                max_value=theta_cut_max,
-            )
+        elif cut_type_theta == "dynamic" or "custom":
+            if cut_type_theta == "dynamic":
```
jsitarek, line 450: here you could do the same with swapping ifs

jsitarek:
thanks a lot @gabemery 
I am merging now

PR #282, opened 2025-02-28
fix list_from_h5 issue

Elisa-Visentin:
Fix issue in LST db query (mixed bool & string format for MAGIC_stereo)

### DIFF
--- a/magicctapipe/scripts/lst1_magic/semi_automatic_scripts/list_from_h5.py
+++ b/magicctapipe/scripts/lst1_magic/semi_automatic_scripts/list_from_h5.py
@@ -256,12 +256,12 @@ def main():
 
     if source_in is None:
         df_LST.query(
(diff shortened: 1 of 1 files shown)


jsitarek (approved):
I approved, but are you sure is the correct fix? I thought that it was supposed to add possibility that a variable is specified as a bool or as a string and this just adds a new condition on HV

Elisa-Visentin:
@jsitarek Yes sorry there are two commits because of an error in copy-paste (I accidentally removed the HV condition)
now the conditions are the ones that should prevent issues with the database
   
            f'MAGIC_trigger=="L3T" & MAGIC_HV=="Nominal" & (MAGIC_stereo == {stereo} | MAGIC_stereo == "{stereo}") & error_code_nsb=="0"',

            f'source=="{source_in}"& MAGIC_trigger=="L3T" & MAGIC_HV=="Nominal" & (MAGIC_stereo == {stereo} | MAGIC_stereo == "{stereo}") & error_code_nsb=="0"',

PR #283, opened 2025-03-05
some tailcut fixes

Elisa-Visentin:


magicctapipe/scripts/lst1_magic/semi_automatic_scripts/config_auto_MCP.yaml
```diff
@@ -24,7 +24,6 @@ data_selection:
 general:
     base_config_file: ''    # path + name to a custom MCP config file. If not provided, the default config.yaml file will be used
```
jsitarek, line 25: the newly introduced field base_db_config_file is still missing in the .rc file

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/LSTnsb.py
```diff
@@ -197,8 +197,25 @@ def main():
     if len(LST_files) == 1:
         logger.info(f"Run {run_number} already processed")
         return
+    config_db = resource_file("database_config.yaml")
```
jsitarek, line 200: if in coincidence script you read the location of this config file from the auto_MCP config file, why also not here?

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/LSTnsb.py
```diff
@@ -197,8 +197,25 @@ def main():
...
+    LST_h5 = config_dict_db["database_paths"]["LST"]
+    LST_key = config_dict_db["database_keys"]["LST"]
+    df_LST = pd.read_hdf(
+        LST_h5,
+        key=LST_key,
+    )
+
+    tailcut = df_LST[df_LST.LST1_run == run_number].iloc[0]["tailcut"]
+    if tailcut == "":
+        return
```
jsitarek, line 216: would be useful to print an error message and return an error code here

jsitarek:
I checked if we are missing any version directories with this:
ls -d /fefs/aswg/data/real/DL1/*/v* | cut -d/ -f8 | sort | uniq -c
```
    424 v0.10
      2 v0.10.12
      1 v0.10.8_test
      1 v0.10G
      2 v0.10_test3
      1 v0.10_tests
      1 v0.10V
    430 v0.9
     14 v0.9G```
but everything except of 0.9 and 0.10 seem to be some minor tests. 0.10.12 I checked in more details - those are files with CatB calibration, but they are also stored in CatB_tailcut84 instead of regular tailcut84, so best not to include them

File: magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/lstchain_version.py [hunk 4]
```diff
@@ -58,8 +64,19 @@ def version_lstchain(df_LST):
                 max_version = lstchain_versions[j]
 
         if max_version is None:
-            raise ValueError("issue with lstchain versions")
-        name = f"/fefs/aswg/data/real/DL1/{date}/{max_version}/tailcut84/dl1_LST-1.Run{run}.h5"
+            print(f"issue with lstchain versions for run {run}\n\n\n")
```
jsitarek, line 67: you could add here that available versions: {version}, allowed versions: {lstchain_versions}

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/nsb_level.py
```diff
@@ -154,26 +154,39 @@ def main():
...
+        tailcut_list = [
+            i.split("/")[-1]
+            for i in glob.glob(f"/fefs/aswg/data/real/DL1/{date}/{max_common}/tailcut*")
+        ]
+
+        for tail in tailcut_list:
+            if os.path.isfile(
+                f"/fefs/aswg/data/real/DL1/{date}/{max_common}/{tail}/dl1_LST-1.Run{run_number}.h5"
+            ):
+                tailcut = str(tail)
```
jsitarek, line 175: again, some priority or error if we end up with the same run in two tailcuts would be useful

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/nsb_level.py
```diff
@@ -154,26 +154,39 @@ def main():
...
 
         bash_scripts(run_number, date, args.config_file, env_name, cluster, lst_config)
 
     print("Process name: nsb")
     print("To check the jobs submitted to the cluster, type: squeue -n nsb")
     list_of_bash_scripts = np.sort(glob.glob("nsb_*_run_*.sh"))
-
-    if len(list_of_bash_scripts) < 1:
+    
+    if len(list_of_bash_scripts) < 0:
```
jsitarek, line 189: why <0 ?
if the list is empty it will be == 0

jsitarek (commented):
I went through the code, and the changes look good to me (I left some small comments)

Elisa-Visentin:
@jsitarek Thank you. I will go through them and fix them (this is still a draft we needed not to get stuck with 2025 data processing, but it still needs some 'fine-tuning' and quite some tests). Nice to know that there are also directories named 'CatB*' instead of the  standard tailcut*. I have to check again carefully the wildcharts I'm using to be sure that I'm not collecting also these 'non-standard' directories 😞

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/config_auto_MCP.yaml
Elisa-Visentin: Now it should be fixed everywhere (sorry, this was a merging problem due to having a PR on a PR related to another PR....)

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/LSTnsb.py
Elisa-Visentin: fixed

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/nsb_level.py
Elisa-Visentin: I now implemented it such that tailcut is a list, we append a value for each taicut directory in which a run is located (for the last available and processable lstchain version), and, if the length of the array is >1, we skip the run and print a message. If len=1, we write the first (only) element in the table

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/LSTnsb.py
Elisa-Visentin: added (error code 4)

magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/lstchain_version.py
```diff
@@ -58,8 +64,19 @@ def version_lstchain(df_LST):
...
+            i.split("/")[-1]
+            for i in glob.glob(
+                f"/fefs/aswg/data/real/DL1/{date}/{max_version}/tailcut*"
+            )
+        ]
+        for tail in tailcut_list:
+            if os.path.isfile(
+                f"/fefs/aswg/data/real/DL1/{date}/{max_version}/{tail}/dl1_LST-1.Run{run}.h5"
+            ):
+                name = f"/fefs/aswg/data/real/DL1/{date}/{max_version}/{tail}/dl1_LST-1.Run{run}.h5"
```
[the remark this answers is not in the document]
Elisa-Visentin: mhhh, maybe not to be put here (since here I only fill the list of available lstchain  versions, and, in case of more than on etailcut, well this version exist), but at the end of the function, where I fill the column with the name of the file available for the last lstchain version (the last in general (e.g., 0.11), not the last we can process (0.10), since we take care of the last processable version in the next script), so that the name of the file is unique and well determined (we don't use it for data processing but just as a human check of what we are processing wrt what we have, but having a name that could change every time we run the script, given to the order of tailcuts in the list over which we loop, could be annoying)

jsitarek (approved):
the changes look good to me, shall we merge?

PR #286, opened 2025-03-28
Implementation of the cloud profile estimator in magic-cta-pipe

mpecimotika:
This script uses geometrical model [(article)](https://doi.org/10.1016/j.jheap.2024.03.003) for the estimation of the vertical transmission, base height and thickness of a cloud, the so-called "proton" LIDAR.

### DIFF
--- a/magicctapipe/resources/config.yaml
+++ b/magicctapipe/resources/config.yaml
@@ -277,11 +277,19 @@ dl2_to_dl3:
     source_name: "Crab"
     source_ra: null  # used when the source name cannot be resolved
     source_dec: null  # used when the source name cannot be resolved
-    
+
+
 cloud_correction:
     specify_cloud: yes
-    base_height: "5900. m" 
+    base_height: "5900. m"
     thickness: "1320. m"
     vertical_transmission: 1.0
     number_of_layers: 10
 
+
+cloud_estimator:
+    quality_cuts: "(impact > 0) & (total_uncert <= 10) & (n_pixels > 20) & ((slope < -5)| (slope > 5)) & (slope < 15) & (slope > -15)"
+    max_offset: "0.2 deg"
+    start_altitude: "3 km"
+    end_altitude: "15 km"
+    n_bins_altitude: 121
--- a/magicctapipe/scripts/lst1_magic/lst1_magic_cloud_profile_estimator.py
+++ b/magicctapipe/scripts/lst1_magic/lst1_magic_cloud_profile_estimator.py
@@ -26,21 +26,21 @@ $ python lst1_magic_cloud_profile_estimator.py
 import argparse
 import logging
 import time
-import magicctapipe
-import ctapipe
 from pathlib import Path
 
 import astropy.units as u
-import tables
+import ctapipe
 import h5py
 import numpy as np
 import pandas as pd
+import tables
 import yaml
-
 from astropy.coordinates import AltAz, SkyCoord
 from ctapipe.coordinates import TelescopeFrame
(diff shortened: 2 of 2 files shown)


magicctapipe/scripts/lst1_magic/lst1_magic_cloud_profile_estimator.py
```diff
@@ -0,0 +1,390 @@
...
+    """
+    Geometrical model needed to estimate the vertical tranmission profile of a cloud.
+    Calculates an oﬀset angle from the primary direction corresponding
+    to the emission height of the Cherenkov light need to construct the longitudinal
+    profile of the observed Cherenkov light
+    and to estimate the vertical tranmission profile of a cloud.
+
+    Parameters
+    ----------
+    imp : float
```
jsitarek, line 60: imp and h are actually not floats but Quantity objects (you can see in line 75 that we convert the ratio to a number without a unit

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_profile_estimator.py
```diff
@@ -0,0 +1,390 @@
...
+
+    Parameters
+    ----------
+    telescopes_images : dict
+        Dictionary containing telescope's details and image.
+    dl1_params : pandas.DataFrame
+        Pandas DataFrame containing DL1 and stereo-reconstruced parameters.
+    config : dict
+        Configuration dictionary containing various processing parameters, including those for
+        the estimation of the longitudinal distribution of the deteced Cherenkov light.
```
jsitarek, line 98: detected

magicctapipe/scripts/lst1_magic/lst1_magic_cloud_profile_estimator.py
```diff
@@ -0,0 +1,390 @@
...
+    assigned_tel_ids : dict
+        Dictionary mapping telescope types to their respective telescope IDs.
+    magic_only_analysis : bool, optional
+        If `True`, the function reconstructs the stereo parameters using only MAGIC events (default is `False`).
+
+    Returns
+    -------
+    pandas.DataFrame
+        Updated dl1_params DataFrame including the longitudinal profile of the observed Cherenkov light.
```
jsitarek, line 107: the original script produced separate output files with the aggregated longitudinal profile (summed up over all the events) from the whole file. 
The way how you did it now with saving a profile for each event is of course more flexible (in terms of applying any additional event selection, because for computing the transmission we either way need statistics of many events), but the output files will be pretty bulky and running over them could be troublesome. It can stay like this for the moment

magicctapipe/resources/config.yaml
```diff
@@ -277,11 +277,19 @@ dl2_to_dl3:
...
     specify_cloud: yes
-    base_height: "5900. m" 
+    base_height: "5900. m"
     thickness: "1320. m"
     vertical_transmission: 1.0
     number_of_layers: 10
 
+
+cloud_estimator:
+    quality_cuts: "(impact > 0) & (total_uncert <= 10) & (n_pixels > 20) & ((slope < -5)| (slope > 5)) & (slope < 15) & (slope > -15)"
```
jsitarek, line 291: I do not remember why I introduced this total_uncert in the first place, but in the config file of the original script I set it to 10 deg which is huge effectively removing this cut

mpecimotika: I agree this is unnecessary cut. I can remove it in the update code.

PR #288, opened 2025-05-20
added a margin of 0.02 in coszd interpolation,

jsitarek:
such that runs that are just outside of the range covered by MCs are processed with "nearest" option


magicctapipe/scripts/lst1_magic/lst1_magic_dl2_to_dl3.py
```diff
@@ -157,6 +158,25 @@ def dl2_to_dl3(input_file_dl2, input_dir_irf, output_dir, config):
     # Prepare for the IRF interpolations
     interpolation_method = config_dl3.pop("interpolation_method")
     logger.info(f"\nInterpolation method: {interpolation_method}")
+    if (scheme == "cosZd") and (interpolation_method != "nearest"):
```
Elisa-Visentin, line 161: I don't know if this could be also used for 'codsZdAz' interpolation.... It is always a linear interpolation, and it fails if the point is outside IRF zenith range

Elisa-Visentin: actually, since both linear and cubic interpolations require 3 points around the real data pointing, probably we could just apply it in case interpolation =/= nearest, w/o caring about the actual (not nearest) interpolation method and scheme? And maybe, as a condition for falling to nearest, we can ask to have at least 3 irfs, so that, in case of tests or schools, if we have only 1 or 2 irfs to produce toy dl3, the script will automatically go for a nearest 'interpolation'?

jsitarek: There are two conditions - first the total number of IRFs is checked (this is a part of the old code), but you can have a situation that you have 10 IRFs, but the target point does not fit in their span. 
I can expand the code to zd-az interpolation, however this one is not what is recommended to be used, and there is some issue of how Az is treated, so we can also go outside of Az range, but well, it would still work better than the original code if somebody applies Az-Zd interpolation

PR #289, opened 2025-05-21
moved some of the parsing of arguments into a separate function

jsitarek:
they were repeated in multiple scripts.

This allows to shorten up the code somewhat


jsitarek:
I saw also that some scripts use a different style of min/max dates. Those I did not touch yet, but it would be good to unify them.

Elisa-Visentin (approved):

Elisa-Visentin:
Fine for me, thank you!

PR #295, opened 2025-08-13
drop duplicates in ST list

Elisa-Visentin:


Elisa-Visentin:
@jsitarek Ok, now there are no duplicates in ST*.txt files (both DL2 and DL3), but still duplicates in list_return (since we run the same runs twice, once for each lstchain version). Should in this case keep the second (i.e., more recent) entries, since in case of an error in first pass but no error in second one we will have a good output file? Or keeping the one (if there is one) without errors, despite, in case of an error in the second pass, the output file could have already been partially overwritten with a corrupted one? Any idea?

jsitarek:
Hi @Elisa-Visentin 

The solution is not perfect, but I cannot find a better one with the limitations that we have currently on the database. We could introduce the LST version number in the naming of ST*.txt files, but this would need to be properly propagated in multiple places, including the job accounting script, and taking into account that we had only a single night like this so far I think it is not worth it, so I am fine with your proposed solution. 

Concerning running those files twice: in the end the file should be exactly the same, because they are run on the same input DL1Stereo file, therefore any failing can happen only due to cluster issues. We would get a problem if both jobs are being run at the same time, but since v0.9 and v0.10 are run separately by @joannaWojS and we have only one night like this, it should be pretty unprobable that this overwritting will happen now that you fixes ST*.txt files. So again, it is probably not worth to complicate the code to avoid double running in this case.

Elisa-Visentin:
Ok, so I will implement the easiest solution (keep last lines in return file), test it and let you know for merging. What I can easily do is also adding a printed message when removing duplicates lines, so that, when running the dl1-dl2 script, we can immediately know the affected date and source, and check those few runs more carefully (e.g., reading all their logs manually). Adding and propagating lstchain version in all the file names and job accounting would be the best solution, but also the most complicated one, and, as we already saw a few days ago, the more loops and ifs we add (period, date, nsb, version...) the more probable is to run into some nasty bugs, and the slower the code. Hopefully, there won't be a lot of data affected by this issue...

PR #299, opened 2025-08-18
New path auto tailcut

Elisa-Visentin:


magicctapipe/scripts/lst1_magic/semi_automatic_scripts/database_production/nsb_level.py
```diff
@@ -134,31 +136,41 @@ def main():
...
+                for i in glob.glob(f"{base_path}/{date}/{max_common}/tailcut*")
+            ]
+
+            for tail in tailcut_list:
+                if os.path.isfile(
+                    f"{base_path}/{date}/{max_common}/{tail}/dl1_LST-1.Run{run_number}.h5"
+                ):
+                    tailcut.append(tail)
+                    name = f"{base_path}/{date}/{max_common}/{tail}/dl1_LST-1.Run{run_number}.h5"
+        if len(np.unique(tailcut)) > 1:
```
jsitarek, line 158: I think this will also fail if you have two copies of the same file in the old and in the new path. I am not sure if this can happen (but I would not be surprised if it does). We could avoid it if we additionally reset tailcut to [] in line 146. In this case if both directories exist it would only care about the new path. 
But if you prefer to keep it as it is, we can also do it.

Elisa-Visentin: It was planned to fail in case we have {old_path}/{last_v}/{tail_1}/Runxx and {new_path}/{last_v}/{tail2}/Runxx  since there are two different tailcuts associated with the same version, same run, and it is not so easy to 'date' them (i.e., find out which is the most recent processed file). But, if we trust OSA not to mess up things (i.e., newer things always go in the new path, and not in the old one), we can fix it to consider only the new path in cases like that: pros: we can recover some runs. And it could happen for archival data (e.g., v0.10) if not already reprocessed with new cuts in old path, but reprocessed with new cuts in new path. Probably it is safer the way you are suggesting (not to discard too many runs), and hoping newer reprocessings of the same run will never go to an older path wrt the one in which the older version is [that would be strange, and hopefully rare/impossible]

Elisa-Visentin: Well, what can happen now is that if there is in the new path a certain date and version, but not all the runs of that date are there (e.g., something has not been reprocessed and is still in old version), it will reset the list but won't append anything new to it (since there is not that file in the new path), so I added a check on that not to try indexing an empty list. Hopefully, also for that, this won't happen (if they reprocess a date, they should reprocess everything and not only some runs)

jsitarek:
thanks for the fixes. You have some CI check formatting issues but once those are fixed and if nothing pops up in the automatic tests (it should not since you only changed autoMCP code) let's merge it.

PR #300, opened 2025-08-19
bug fix

Elisa-Visentin:


Elisa-Visentin:
Wait, there is an issue with the new dl3 format. the events hdu must be called on a table where gammaness cuts are already implemented. Ugh

PR #328, opened 2026-04-03
changes to use get_stereo_events everywhere instead of get_stereo_events_old

jsitarek:
in some places in the code the old function was still used (which was not not compatible with more LSTs).

some things to be fixed still
DL1 to DL2 step needs now a config file. In the basic script this is implemented, but needs to be included also in the autoMCP scripts

the naming of the telescopes is different LST-1 instead of LST1, MAGIC-I, MAGIC-II instead of M1 M2, and those names are used in the RF file names. In order to avoid rerunning/renaming the files we should look also for the files with the old names.

### DIFF
--- a/magicctapipe/io/io.py
+++ b/magicctapipe/io/io.py
@@ -783,7 +783,7 @@ def load_magic_dl1_data_files(input_dir, config):
 
 
 def load_train_data_files(
-    input_dir, offaxis_min=None, offaxis_max=None, true_event_class=None
+    input_dir, config, offaxis_min=None, offaxis_max=None, true_event_class=None
 ):
     """
     Loads DL1-stereo data files and separates the shower events per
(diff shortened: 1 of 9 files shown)


Elisa-Visentin:
@jsitarek, @aleberti  4 LSts were implemented up to Dl1 stereo (thus keeping also the get_astereo_events_old function) since there was a never-ended discussion about Rfs (one per telescope  vs one per combination, the former as the best for 4-6 tels, the latter the best one for MAGIC+LST1), thus keeping in separate branches the DL1 Stereo to Dl3 multi-LSTs version. Since the other LSTs will hopefulyl be ready soon, I think we can try discussing that during one of our meetings, and try to merge all these separate branches (maybe adding a flag in the train and dl1-dl2 script to choose in between per tel and per combo on the fly, thus also solving the naming issues and being able to still use our whole MC library?) into a v0.7 allowing to deal with all the LSTs up to DL3, for x-checks wrt the official LSTs sw, and for still allowing MAGIC+LSTs (studies of the performance of different setups are ongoing, and they are now using a draft branch I am trying to keep updated with all the ctapipe and lstchain updates, but having it in the master would be easier) [ok, this comment is a bit messy, but the idea is: let's try to discuss that during a meeting, so that we can set up a list of everything we should still fix for this 4LSTs compatibility]

PR #329, opened 2026-04-07
V0.6.1+4lst

Elisa-Visentin:


Elisa-Visentin:
There are probably some tests not actually needed, but we can start to have a look into that to start pruning the 'tree' of 4 LST PRs

magicctapipe/io/tests/test_io.py
```diff
@@ -21,12 +23,77 @@
     load_mc_dl2_data_file,
     load_train_data_files,
     load_train_data_files_tel,
+    query_data,
     save_pandas_data_in_table,
     telescope_combinations,
 )
 
+LOGGER = logging.getLogger(__name__)
```
jsitarek, line 31: this LOGGER does not seem to be used anywhere

magicctapipe/io/tests/test_io.py
```diff
@@ -1,11 +1,13 @@
 import glob
+import logging
```
jsitarek, line 2: in principle the tests are by default executed with all the text output supressed, do we need this part?

magicctapipe/io/tests/test_io.py
```diff
@@ -21,12 +23,77 @@
...
+        data = query_data(query_test, "software", 3, [], False)
+        assert np.allclose(
+            np.array(data["event_id"]), np.array([0, 3, 4, 5, 6, 7, 9, 10, 11])
+        )
+
+        with caplog.at_level(logging.WARNING):
+            query_test = pd.read_hdf(query_test_1, key="/events/parameters")
+            data = query_data(query_test, "software", None, [], False)
+        assert (
+            'Requested event type and provided telescopes IDs are not consistent; "software" must be used in case of standard MAGIC+LST-1 analyses'
```
jsitarek, line 51: now I see where you use the log, but such a test is dangerous: enough that somebody will change slightly the message and the test will fail. This looks like a fatal error, so we could e.g. throw an exception that can be tested here

magicctapipe/io/tests/test_io.py
```diff
@@ -21,12 +23,77 @@
...
+        data = query_data(query_test, "trigger_no_magic_stereo", 3, [], False)
+        assert np.allclose(
+            np.array(data["event_id"]), np.array([0, 1, 3, 4, 6, 7, 9, 10, 11])
+        )
+
+        with caplog.at_level(logging.WARNING):
+            query_test = pd.read_hdf(query_test_2, key="/events/parameters")
+            data = query_data(query_test, "magic_only", None, [], False)
+        assert (
+            "MAGIC-only analysis requested, but inconsistent with the provided telescope IDs: check the configuration file"
```
jsitarek, line 75: as above - better to e.g. throw an exception that is tested here

magicctapipe/io/tests/test_io.py
```diff
@@ -21,12 +23,77 @@
...
+            in caplog.text
+        )
+
+        query_test = pd.read_hdf(query_test_2, key="/events/parameters")
+        data = query_data(query_test, "magic_only", 3, [], False)
+        assert np.allclose(np.array(data["event_id"]), np.array([2, 5, 8]))
+
+        query_test = pd.read_hdf(query_test_2, key="/events/parameters")
```
jsitarek, line 83: since there are many lines like this it can be passed as parametrized test or just a loop with input/output values listed in an array/dictionaries

magicctapipe/io/tests/test_io.py
```diff
@@ -21,12 +23,77 @@
...
+        assert (
+            'Requested event type and provided telescopes IDs are not consistent; "software" must be used in case of standard MAGIC+LST-1 analyses'
+            in caplog.text
+        )
+
+        query_test = pd.read_hdf(query_test_2, key="/events/parameters")
+        data = query_data(query_test, "trigger_3tels_or_more", None, [4, 5, 6], False)
+        assert np.allclose(np.array(data["event_id"]), np.array([1, 3, 9]))
+
+        query_test = pd.read_hdf(query_test_2, key="/events/parameters")
```
jsitarek, line 59: this line (and e.g. in L65) do not seem needed for me since query_test is not expected to change from L55

magicctapipe/io/tests/test_io.py
```diff
@@ -199,11 +266,17 @@ def test_telescope_combinations(self, config_gen, config_gen_4lst):
...
-            "LST-1_MAGIC-I": [1, 2],
-            "LST-1_MAGIC-I_MAGIC-II": [1, 2, 3],
-            "LST-1_MAGIC-II": [1, 3],
-            "MAGIC-I_MAGIC-II": [2, 3],
+            "LST-1_MAGIC-I": [1, 2],  # combo_type = 0
+            "LST-1_MAGIC-I_MAGIC-II": [1, 2, 3],  # combo_type = 1
+            "LST-1_MAGIC-II": [1, 3],  # combo_type = 2
+            "MAGIC-I_MAGIC-II": [2, 3],  # combo_type = 3
         }
+        assert list(M_LST_comb.keys()) == [
```
jsitarek, line 274: this one is not needed, if you check the whole dictionary the test above will fail already if there is a difference in the keys

magicctapipe/io/tests/test_io.py
```diff
@@ -574,6 +778,71 @@ def test_load_irf_files_exc(self, temp_irf_exc):
...
+        """
+
+        irf, header = load_irf_files(str(IRF_tel))
+        assert set(list(irf.keys())).issubset(
+            set(
+                [
+                    "grid_points",
+                    "effective_area",
+                    "energy_dispersion",
+                    "psf_table",
```
jsitarek, line 795: instead of checking later on that psf_table is not in irf keys, you could simply remove it from here, same with background

magicctapipe/io/tests/test_io_monly.py
```diff
@@ -142,6 +147,15 @@ def test_exist_rf(RF_monly):
     """
 
     assert len(glob.glob(f"{RF_monly}/*")) == 3
+    assert len(glob.glob(f"{RF_monly}/energy*MAGIC-I_MAGIC-II_unsigned.joblib")) == 1
+    assert len(glob.glob(f"{RF_monly}/disp*MAGIC-I_MAGIC-II_unsigned.joblib")) == 1
+    assert len(glob.glob(f"{RF_monly}/event*MAGIC-I_MAGIC-II_unsigned.joblib")) == 1
+    assert len(glob.glob(f"{RF_monly}/energy*LST-1*_unsigned.joblib")) == 0
```
jsitarek, line 153: you could make a double loop over "energy_regressors", "disp_regressors", "event_classifiers" and over telescope types (then you would be able to skip the first '*')

magicctapipe/io/tests/test_io_monly.py
```diff
@@ -154,9 +168,39 @@ def test_exist_dl2_mc(p_dl2_monly, gamma_dl2_monly):
...
+    assert len(glob.glob(f"{RF_monly_tel}/*")) == 6
+    assert (
+        len(glob.glob(f"{RF_monly_tel}/energy*MAGIC-I_MAGIC-II_unsigned.joblib")) == 0
+    )
+    assert len(glob.glob(f"{RF_monly_tel}/disp*MAGIC-I_MAGIC-II_unsigned.joblib")) == 0
+    assert len(glob.glob(f"{RF_monly_tel}/event*MAGIC-I_MAGIC-II_unsigned.joblib")) == 0
+    assert len(glob.glob(f"{RF_monly_tel}/energy*LST-1*_unsigned.joblib")) == 0
+    assert len(glob.glob(f"{RF_monly_tel}/disp*LST-1*_unsigned.joblib")) == 0
+    assert len(glob.glob(f"{RF_monly_tel}/event*LST-1*_unsigned.joblib")) == 0
+    assert len(glob.glob(f"{RF_monly_tel}/energy*MAGIC-I_unsigned.joblib")) == 1
```
jsitarek, line 186: you can also use loop here,  
 the remaining 3 files are not checked

magicctapipe/io/tests/test_io_monly.py
```diff
@@ -262,6 +306,112 @@ def test_get_dl2_mean_exc(self, p_dl2_monly, gamma_dl2_monly):
...
+            assert np.all(data["gammaness"] > 0.1)
+            assert len(data) > 0
+
+    def test_load_mc_dl2_data_file_opt_tel(
+        self, config_gen, p_dl2_monly_tel, gamma_dl2_monly_tel
+    ):
+        """
+        Check on event_type
+        """
+        dl2_mc = [p for p in gamma_dl2_monly_tel.glob("*")] + [
```
jsitarek, line 353: this is used in a bunch of functions. maybe you can turn it e.g. into fixture

magicctapipe/io/tests/test_io_monly.py
```diff
@@ -271,6 +421,15 @@ def test_exist_irf(IRF_monly):
     assert len(glob.glob(f"{IRF_monly}/*")) == 1
 
 
+@pytest.mark.dependency(depends=["test_exist_dl2_mc_tel"])
+def test_exist_irf_tel(IRF_monly_tel):
+    """
+    Check if IRFs produced
+    """
+
+    assert len(glob.glob(f"{IRF_monly_tel}/*")) == 1
```
jsitarek, line 430: you could specify a bit more the file format

magicctapipe/io/io.py
```diff
@@ -73,6 +73,73 @@
...
+    event_type : str
+        Type of events to be selected
+        'software' : Stadard MAGIC+LST-1 case; remove magic-only events
+        'trigger_3tels_or_more' : at least three telescopes triggered
+        'trigger_no_magic_stereo' : no need for both MAGIC to trigger, to be used if more than one LST is used; remove MAGIC-only data if they exist
+        'magic_only' : only M1_M2 events selected
+        'hardware' : hardware trigger
+    magic_only : int or None
+        Index of the M1_M2 combination (if it exist, otherwise it is equal to None)
+    three_or_more : list
```
jsitarek, line 94: it is not very natural passing this list to the function. If user selects 3 or more the list should be automatically created inside the function (it does not take much time). Similarly magic_only combination should be also computed inside the function -  pass the config file as an argument go get needed tel numbers

magicctapipe/io/io.py
```diff
@@ -1125,33 +1114,29 @@ def load_dl2_data_file(input_file, quality_cuts, event_type, weight_type_dl2):
         If the input event type is not known
     """
 
+    _, TEL_COMBINATIONS = telescope_combinations(config)
+    three_or_more = []
```
jsitarek, line 1118: this part is repeated in two functions - one more reason to move it to query_data

magicctapipe/resources/config.yaml
```diff
@@ -1,18 +1,18 @@
 mc_tel_ids:
     LST-1: 1
-    LST-2: 0
```
jsitarek, line 3: the default config should not be changed at the current moment

magicctapipe/scripts/lst1_magic/lst1_magic_dl1_stereo_to_dl2.py
```diff
@@ -260,8 +272,12 @@ def dl1_stereo_to_dl2(input_file_dl1, input_dir_rfs, output_dir):
         Path to a directory where trained RFs are stored
     output_dir : str
         Path to a directory where to save an output DL2 data file
+    config : dict
+        Configuration for the LST-1 + MAGIC analysis
+    reco_combo : bool
+        If True, reconstructs using RF per telescope per telescope combination, if False, use RF per telescope
     """
-
+    TEL_NAMES, _ = telescope_combinations(config)
```
jsitarek, line 280: I guess we will be able to use the old RFs without renaming simply modifying the name of the telescope in the config, have you tried it?

magicctapipe/scripts/lst1_magic/lst1_magic_train_rfs.py
```diff
@@ -141,41 +151,65 @@ def train_energy_regressor(input_dir, output_dir, config, use_unsigned_features=
     logger.info(format_object(config_rf["features"]))
 
     logger.info(f"\nUse unsigned features: {use_unsigned_features}")
-
-    energy_regressor = EnergyRegressor(
-        config_rf["settings"], config_rf["features"], use_unsigned_features
-    )
+    logger.info(f"\nconfiguration file: {config}")
+    logger.info(f'\nmc_tel_ids: {config["mc_tel_ids"]}')
+    if train_combo:
```
jsitarek, line 156: both part of the if condition execute the same code

magicctapipe/scripts/lst1_magic/lst1_magic_train_rfs.py
```diff
@@ -213,42 +255,64 @@ def train_disp_regressor(input_dir, output_dir, config, use_unsigned_features=Fa
     logger.info(format_object(config_rf["features"]))
 
     logger.info(f"\nUse unsigned features: {use_unsigned_features}")
-
-    disp_regressor = DispRegressor(
-        config_rf["settings"], config_rf["features"], use_unsigned_features
-    )
+    if train_combo:
+        disp_regressor = DispRegressor(
```
jsitarek, line 259: same as in the energy one - both parts of the condition execute the same code

magicctapipe/scripts/lst1_magic/lst1_magic_train_rfs.py
```diff
@@ -296,22 +379,31 @@ def train_event_classifier(
     logger.info(format_object(config_rf["features"]))
 
     logger.info(f"\nUse unsigned features: {use_unsigned_features}")
-
-    event_classifier = EventClassifier(
-        config_rf["settings"], config_rf["features"], use_unsigned_features
-    )
+    if train_combo:
+        event_classifier = EventClassifier(
+            TEL_NAMES,
```
jsitarek, line 384: and also here it is the same

magicctapipe/conftest.py
```diff
@@ -976,13 +1362,34 @@ def real_dl2(coincidence_stereo, RF, temp_DL2_real):
...
             ]
         )
     return temp_DL2_real
 
 
 @pytest.fixture(scope="session")
-def real_dl2_monly(stereo_monly, RF_monly, temp_DL2_real_monly):
+def real_dl2_tel(coincidence_stereo, RF_tel, temp_DL2_real_tel, config):
+    """
+    Produce a DL2 file
```
jsitarek, line 1375: since they are now in different flavours you can add in the docstring if they are from training one per telescope or one per combo_type

magicctapipe/io/tests/test_io.py
Elisa-Visentin: removed, trying to fix that in io.py raising an exception together with the warning message

magicctapipe/io/tests/test_io.py
Elisa-Visentin: as above

magicctapipe/io/tests/test_io.py
Elisa-Visentin: same

magicctapipe/io/tests/test_io.py
Elisa-Visentin: same

PR #330, opened 2026-04-09
Update .pre-commit-config.yaml

Elisa-Visentin:


Elisa-Visentin:
@jsitarek, @aleberti. This seems to solve the recent issue with flake8. If there is any other better solution, please let me know :)

jsitarek:
thanks a lot @Elisa-Visentin 
looks like you needed to increase the version number, which seems to me to be a very good solution.
if @aleberti also agrees, let's merge it.

aleberti:
@Elisa-Visentin @jsitarek fine for me!

(this week I am on vacation, so that's why I am a bit unrespsonsive)

jsitarek (approved):

